PR #13, opened 2026-05-09
refactor: workspace setup + GPU selection abstraction + NVAPI/NVML dual-path (closes #2, #4)

### LINKED ISSUE #2: 使用 workspace 管理依赖
希望 Python 和 Rust 都用 workspace 模式管理依赖。

看起来大家普遍喜爱 uv，用 uv 管理 Python 项目的 workspace 罢

Dixiao-L:
## Summary

Addresses Issues [#2](../../issues/2) and [#4](../../issues/4) with root-cause fixes — no band-aids.

---

### Issue #2 — Workspace management

- **Rust workspace**: Add root `Cargo.toml` declaring `auto-optimizer` and `srv` as workspace members. Move `[profile.release]` from `auto-optimizer/Cargo.toml` to the workspace root (eliminates per-member profile warnings from Cargo).
- **Python workspace**: Add root `pyproject.toml` as a `uv` workspace, listing `gui`, `tui`, `cli-stressor-cuda`, and `cli-stressor-opencl` as members. Running `uv sync` from the repo root now installs all Python projects into a shared virtual environment.

---

### Issue #4 — GPU enumeration architecture + NVAPI/NVML dual-path

**Root cause**: GPU selection was coupled to `clap::parser::ValuesRef` (a CLI-only type), making library-style calls impossible. Additionally, `info` and `status` used NVAPI exclusively, so server GPUs on Windows (where NVAPI cannot initialize) returned errors for even basic queries.

**Changes**:

#### `GpuSelector` — decouples GPU selection from `clap`

```rust
// Before: every handler had this signature fragment
gpu: Option<clap::parser::ValuesRef<'_, String>>

// After: library-friendly type that works from CLI or direct code
let selector = GpuSelector::from_clap(matches.get_many("gpu")); // CLI path
let selector = GpuSelector::from_specs(vec!["0".into()]);       // library path
let selector = GpuSelector::all();                               // select all
```

#### Pre-selection in `main.rs` — outer layer owns GPU selection

GPU enumeration and selection now happen **once** in `main.rs` before any subcommand dispatch. Handlers receive `&[&Gpu]` / `&[u32]` slices of already-selected GPUs and no longer call `select_gpus` internally.

```
Before:  main → handle_info(ValuesRef) → get_sorted_gpus() → select_gpus()
After:   main → enumerate → select → handle_info(&[&Gpu], Option<&Nvml>, &[u32])
```

#### NVAPI/NVML dual-path for `info` and `status`

`handle_info` and `handle_status` now:
1. Use the NVAPI path when pre-selected NVAPI GPUs are available (existing behaviour, unchanged)
2. Fall back to NVML when NVAPI is unavailable — covers server/professional GPUs on Windows

NVML `info` output: name, PCI bus/device/domain/IDs, UUID, VBIOS version  
NVML `status` output: temperature, core/mem clocks, power, fan speed, GPU/mem utilization, VRAM usage

---

## Test plan

- [ ] `cargo check` passes with no errors (one dead-code warning for `GpuSelector::all` / `from_specs` as expected — these are library API surface, unused only in the binary)
- [ ] `cargo build --release` on Windows with a consumer GPU (NVAPI + NVML) — existing `info`/`status`/`get`/`reset`/`set` commands behave identically
- [ ] On a Windows server GPU (NVAPI unavailable): `nvoc info` and `nvoc status` now succeed via NVML fallback instead of erroring out
- [ ] `uv sync` from repo root installs all four Python projects
- [ ] `cargo build` from repo root builds both `auto-optimizer` and `srv`

https://claude.ai/code/session_01VwbR1hGQXZuVrtuG2NUa6Z

---
_Generated by [Claude Code](https://claude.ai/code/session_01VwbR1hGQXZuVrtuG2NUa6Z)_


physicsdolphin (approved):

Dixiao-L:
Reviewed the diff alongside PR #14 — both are open against `main` and target the same architectural issues (#2 + #4). Sharing observations rather than blocking; the maintainers should pick a merge order.

## Overlap with PR #14

Both PRs modify three files:
- `auto-optimizer/src/basic_func.rs` — this PR: +323/-170, PR #14: +264/-68
- `auto-optimizer/src/main.rs` — this PR: +60/-42, PR #14: +101/-47
- `auto-optimizer/src/oc_get_set_function_nvapi.rs` — this PR: +5/-5, PR #14: +9/-7

The two PRs will conflict heavily on `basic_func.rs` and `main.rs`. They take different shapes:
- This PR introduces a `GpuSelector` type that wraps `Option<Vec<String>>` and exposes `from_clap`, `from_specs`, `all`. Library callers can construct selectors without going through clap.
- PR #14 keeps the `Option<&[String]>` slice signature and threads `&Nvml` through the call graph instead.

Both are reasonable. `GpuSelector` is the more library-friendly long-term API (matches the spirit of #3); the slice approach in PR #14 is mechanically smaller. Worth picking one before merging the other to avoid a painful rebase.

PR #14 also has a near-total rewrite of `oc_get_set_function_nvml.rs` (+265/-494) that this PR does not touch. That's the *bigger* lift toward issue #4. Consider merging this PR first (workspace + GpuSelector) and rebasing PR #14 on top.

## A bug from #47 still present in `select_gpus`

Issue #47 is the silent-drop in `select_gpus`/`select_gpu_ids` when a `--gpu` value doesn't match any device. The new code in this PR keeps the silent-skip behavior:

```rust
// auto-optimizer/src/basic_func.rs (after this PR's diff)
for input in inputs {
    if input < 256 {
        if let Some(g) = gpus.get(input) {
            result.push(g);
            continue;
        } else {
            continue;          // ← still silently drops
        }
    }
    if let Some(g) = gpus.iter().find(|g| g.id() == input) {
        result.push(g);
        continue;
    }
    let legacy = input << 8;
    if let Some(g) = gpus.iter().find(|g| g.id() == legacy) {
        result.push(g);
        continue;
    }
    // No match — silently dropped
}
```

`--gpu 0,99` on a 1-GPU system still returns `Ok([gpu_0])` with no warning that GPU 99 was ignored. This was the core complaint of #47. The signature change to `&GpuSelector` is orthogonal to fixing the silent-drop; a follow-up commit on this branch could:

```rust
match matched {
    Some(g) => result.push(g),
    None => return Err(Error::Custom(format!("no GPU matches --gpu {}", input))),
}
```

(or emit a `eprintln!` warning if the soft-skip is preferred).

If you'd like, this can fold in here while the PR is in review; otherwise it can be done in a follow-up after merge.

## Smaller observations

- The `pub struct GpuSelector(Option<Vec<String>>);` newtype is good. Consider deriving `Debug, Clone` and adding a `From<Vec<String>>` impl for ergonomics (would let library callers write `GpuSelector::from(vec!["0".into()])`).
- `GpuSelector::from_specs(impl IntoIterator<Item = String>)` is a clean signature; `&'a str` impl might also be welcome to avoid forcing callers to allocate.
- The new NVML-fallback paths in `handle_info`/`handle_status` look good. One nit: `if !nvapi_gpus.is_empty()` is the gate, but a multi-GPU machine with mixed availability (some NvAPI-visible, some only NVML-visible) would still fall fully into the NVAPI branch. Probably fine for now; flag it for the eventual #4 follow-up.

Not blocking. The structural direction is correct and matches issue #4's plan. Just want to make sure #47 doesn't slip through the cracks.

---

PR #21, opened 2026-05-09
fix(auto-optimizer): prevent u32 overflow in set_legacy_clocks_nvapi before unsafe FFI (closes #16)

### LINKED ISSUE #16: auto-optimizer: `set_legacy_clocks_nvapi` overflows on user MHz input — no clamp before unsafe FFI
## Summary

`set_legacy_clocks_nvapi` multiplies the user-supplied MHz values by fixed scale factors (1000 for memory, 2000 for core) without any upstream validation, range check, or saturating arithmetic. Sufficiently large input values overflow `u32` — panic in debug builds, silent wrap in release builds — and write garbage into the `NvClocksInfo` struct that's then handed to the undocumented `NvAPI_GPU_SetClocks` legacy interface.

## Evidence

The write itself, `auto-optimizer/src/oc_get_set_function_nvapi.rs:1480-1538`:

```rust
pub fn set_legacy_clocks_nvapi(gpu: &Gpu, core_mhz: u32, mem_mhz: u32) -> Result<(), Error> {
    use nvapi_hi::sys::nvapi_QueryInterface;
    use std::mem;

    const NVAPI_GPU_SET_CLOCKS_ID: u32 = 0x6f151055;

    #[repr(C)]
    struct NvClocksInfo { version: u32, clocks: [u32; 32] }

    let version = (size_of::<NvClocksInfo>() as u32) | (2 << 16);
    let mut info = NvClocksInfo { version, clocks: [0; 32] };

    info.clocks[8]  = mem_mhz  * 1000;     // ← overflows for mem_mhz  > 4_294_967
    info.clocks[30] = core_mhz * 2000;     // ← overflows for core_mhz > 2_147_483

    unsafe {
        let ptr_res = nvapi_QueryInterface(NVAPI_GPU_SET_CLOCKS_ID);
        let ptr = match ptr_res { Ok(p) => p as *const (), Err(_) => return Err(...) };
        type SetClocksFn = unsafe extern "system" fn(...) -> nvapi_hi::sys::Status;
        let func: SetClocksFn = mem::transmute(ptr);
        let status = func(*gpu.inner().handle(), &mut info);
        ...
    }
}
```

The upstream parse, `auto-optimizer/src/main.rs:151-154`:

```rust
let core_mhz = matches.get_one::<String>("core").unwrap().parse::<u32>()
    .map_err(|_| "Invalid integer for core frequency")?;
let mem_mhz = matches.get_one::<String>("memory").unwrap().parse::<u32>()
    .map_err(|_| "Invalid integer for memory frequency")?;
```

The parser accepts anything up to `u32::MAX`. There is no clamp. The user supplies MHz; the function silently translates to a hardware unit through an unchecked multiplication and hands the result to an undocumented driver interface.

Concrete failure paths:

| Input `--core` | `info.clocks[30]` (release build) | What happens |
|---|---|---|
| `2000` (normal) | `4_000_000` | works as intended |
| `2_147_484` | `200` (wrapped) | a driver call asks the hardware for 200 of whatever-unit, value is meaningless |
| `4_000_000_000` | UB / panic in debug | – |

## Why it matters

This is the *legacy clock interface* — the path used for Maxwell-and-earlier GPUs through an undocumented NvAPI ID extracted from old reverse-engineering tables (`NVAPI_GPU_SET_CLOCKS_ID = 0x6f151055`, comment block on lines 1484-1497 confirms). Drivers in this region are exactly the ones least likely to validate inputs cleanly. A wrap-then-write is the worst case for "crash the driver" debugging because the user typed a number that *looks* small (e.g. `--core 5000000` thinking it's clearly out of range, expecting a clean rejection) and instead gets a hard hang from the driver receiving a tiny wrapped value with the right struct version.

The risk is amplified by the surrounding code: the `unsafe` block trusts the function pointer it `transmute`s, the structure `version` is partly guessed (the comment says *"在此尝试版本 1 和 2 对应的魔力头"* — "try magic headers for v1 and v2"). Any garbage going into this path can crash the driver.

## Recommended approach

Two layers, both small:

1. **Saturating multiply at the call site:**
   ```rust
   info.clocks[8]  = mem_mhz.saturating_mul(1000);
   info.clocks[30] = core_mhz.saturating_mul(2000);
   ```
   This kills the silent wrap without changing happy-path behavior.

2. **Range check at parse time** in `main.rs` (`legacy-clock` arm):
   ```rust
   const MAX_LEGACY_MHZ: u32 = 5_000;     // pick a generous upper bound
   const MIN_LEGACY_MHZ: u32 = 100;
   if !(MIN_LEGACY_MHZ..=MAX_LEGACY_MHZ).contains(&core_mhz) {
       return Err(format!("--core {} MHz out of supported legacy range {}..={}",
                          core_mhz, MIN_LEGACY_MHZ, MAX_LEGACY_MHZ).into());
   }
   ```
   Same for `mem_mhz`. Picking exact bounds is a small follow-up; even a loose `<= 10_000` is dramatically better than no check.

## Severity

Medium — overflow only triggers on user input far outside any real GPU clock range, but when it does, the failure mode is "driver gets a deliberately-corrupt control struct via an undocumented interface", which is the worst class of failure for this codebase.

## Out of scope

- Auditing the rest of the `unsafe` block (the `transmute`d function pointer, the guessed struct version) — that belongs to a wider audit of the undocumented NvAPI surface.
- Adding range checks to other write paths (`set_pstate_base_voltage`, the VFP curve setters); same pattern likely applies and could be a follow-up sweep.

Dixiao-L:
## Summary

Fixes **issue #16** — `set_legacy_clocks_nvapi` could silently wrap (release) or panic (debug) on user-supplied MHz values before passing the corrupted `NvClocksInfo` struct to an undocumented NvAPI legacy interface.

Two-layer fix, both changes are minimal and surgical:

### Layer 1 — Saturating multiply at the FFI call site (`oc_get_set_function_nvapi.rs`)

```rust
// Before (overflows for mem_mhz > 4_294_967 / core_mhz > 2_147_483)
info.clocks[8]  = mem_mhz  * 1000;
info.clocks[30] = core_mhz * 2000;

// After
info.clocks[8]  = mem_mhz.saturating_mul(1000);
info.clocks[30] = core_mhz.saturating_mul(2000);
```

`saturating_mul` clamps to `u32::MAX` instead of wrapping, eliminating the silent corruption path on the hot path immediately before the `unsafe` block. Happy-path behavior is unchanged.

### Layer 2 — Range validation at parse time (`main.rs`, `legacy-clock` arm)

```rust
const MIN_LEGACY_MHZ: u32 = 100;
const MAX_LEGACY_MHZ: u32 = 5_000;
if !(MIN_LEGACY_MHZ..=MAX_LEGACY_MHZ).contains(&core_mhz) {
    return Err(format!("--core {} MHz is outside the supported legacy range {}..={}",
                       core_mhz, MIN_LEGACY_MHZ, MAX_LEGACY_MHZ).into());
}
// same for mem_mhz
```

Out-of-range inputs are rejected with a clear error message before they reach the FFI call site at all. The bounds (100–5000 MHz) are generous and cover any real Maxwell-era GPU clock.

## Root cause

The overflow was inherent in using `*` on `u32` values that the user fully controls. The scale factors (`×1000`, `×2000`) overflow at inputs that look plausibly large (`--core 5000000`) but are still valid `u32` values — exactly the case where a user would expect a clean error and instead gets a wrapped value sent to a fragile undocumented driver interface.

## Test plan

- [ ] `cargo check` passes with no new warnings
- [ ] `cargo build --release` succeeds
- [ ] `nvoc set nvapi legacy-clock --core 2000 --memory 4000` works as before
- [ ] `nvoc set nvapi legacy-clock --core 99` returns error: `--core 99 MHz is outside the supported legacy range 100..=5000`
- [ ] `nvoc set nvapi legacy-clock --core 5001` returns error: `--core 5001 MHz is outside the supported legacy range 100..=5000`
- [ ] `nvoc set nvapi legacy-clock --core 2147484` no longer silently wraps (blocked by range check)

## Out of scope

Per issue #16, a wider audit of the `transmute`d function pointer and guessed struct version in the same `unsafe` block is a separate follow-up.

https://claude.ai/code/session_0171xUTc3XAhuW1x7zBkszSo

---
_Generated by [Claude Code](https://claude.ai/code/session_0171xUTc3XAhuW1x7zBkszSo)_

### DIFF
--- a/auto-optimizer/src/main.rs
+++ b/auto-optimizer/src/main.rs
@@ -152,6 +152,22 @@ fn main_result() -> Result<i32, Box<dyn std::error::Error>> {
                                 .map_err(|_| "Invalid integer for core frequency")?;
                             let mem_mhz = matches.get_one::<String>("memory").unwrap().parse::<u32>()
                                 .map_err(|_| "Invalid integer for memory frequency")?;
+
+                            const MIN_LEGACY_MHZ: u32 = 100;
+                            const MAX_LEGACY_MHZ: u32 = 5_000;
+                            if !(MIN_LEGACY_MHZ..=MAX_LEGACY_MHZ).contains(&core_mhz) {
+                                return Err(format!(
+                                    "--core {} MHz is outside the supported legacy range {}..={}",
+                                    core_mhz, MIN_LEGACY_MHZ, MAX_LEGACY_MHZ
+                                ).into());
+                            }
+                            if !(MIN_LEGACY_MHZ..=MAX_LEGACY_MHZ).contains(&mem_mhz) {
+                                return Err(format!(
+                                    "--memory {} MHz is outside the supported legacy range {}..={}",
+                                    mem_mhz, MIN_LEGACY_MHZ, MAX_LEGACY_MHZ
+                                ).into());
+                            }
+
                             for gpu in &gpus {
                                 match set_legacy_clocks_nvapi(gpu, core_mhz, mem_mhz) {
                                     Ok(_) => println!("Legacy clock applied to GPU: Core = {} MHz, Mem = {} MHz", core_mhz, mem_mhz),
--- a/auto-optimizer/src/oc_get_set_function_nvapi.rs
+++ b/auto-optimizer/src/oc_get_set_function_nvapi.rs
@@ -1502,8 +1502,9 @@ pub fn set_legacy_clocks_nvapi(gpu: &Gpu, core_mhz: u32, mem_mhz: u32) -> Result
     };
 
     // 结合以前的逆向记录进行反向运算，乘以倍率使其满足旧结构格式要求
-    info.clocks[8] = mem_mhz * 1000;
-    info.clocks[30] = core_mhz * 2000;
+    // saturating_mul prevents silent wrap-around overflow before the unsafe FFI call
+    info.clocks[8] = mem_mhz.saturating_mul(1000);
+    info.clocks[30] = core_mhz.saturating_mul(2000);
 
     unsafe {
         // 1. 获取隐藏函数指针


Dixiao-L:
Closing as duplicate. Superseded by #25 (functionally identical diff, more thorough comments).

---

PR #26, opened 2026-05-09
fix(auto-optimizer): replace SystemTime with Instant for fluctuation pacing

### LINKED ISSUE #18: auto-optimizer: autoscan fluctuation timer uses `SystemTime` (wall-clock); clock skew breaks pacing
## Summary

The autoscan inner loop paces a periodic "freq fluctuation" trigger using `SystemTime::now()` and `SystemTime::elapsed()`. `SystemTime` is a wall-clock that can move backward (NTP corrections, manual clock changes, DST jumps on some platforms), and `elapsed()` returns `Err` whenever it does. The current code substitutes a 6-second fallback duration on error, which is *larger* than the 1.5-second trigger threshold — so any clock slip causes the fluctuation step to fire on every loop iteration until `SystemTime::now()` recovers.

`std::time::Instant` is the correct primitive for this kind of pacing because it is monotonic and infallible.

## Evidence

`auto-optimizer/src/oc_scanner.rs:228-298`:

```rust
let test_start_at = Instant::now();          // L228 — uses Instant correctly here
let mut last_fluctuation = SystemTime::now();// L229 — uses SystemTime here
let mut in_test_check_number = 0;
...

loop {
    if last_fluctuation.elapsed().unwrap_or(Duration::from_secs(6))   // L246
        >= Duration::from_millis(1500)
    {
        in_test_check_number += 1;
        println!("inducing freq fluctuation...");

        fluctuation_h_l_flag = apply_fluctuation(gpu, cfg, fluctuation_h_l_flag);
        ...
        last_fluctuation = SystemTime::now();    // L298
    }

    sleep(time::Duration::from_secs(1));
    ...
}
```

Two issues in three lines:

1. **Wrong clock type.** `SystemTime` is a wall-clock; `Instant` is monotonic. Inside the same function `Instant::now()` is already used for `test_start_at` (L228) — the choice on L229 is inconsistent with the surrounding code.
2. **Fallback inverts the safety property.** When `last_fluctuation.elapsed()` returns `Err`, the `unwrap_or(Duration::from_secs(6))` substitutes 6 seconds. Six seconds is *greater than* the 1.5-second trigger, so the `if` body fires. After firing, `last_fluctuation = SystemTime::now()` is called — which may again be in the past relative to the original. The next iteration re-fails, fires again. This continues until the clock catches up. During this window the fluctuation step runs on every 1-second tick instead of every 1.5 seconds, hammering the GPU with extra voltage/freq toggles and skewing the in-test thrm/pwr ratio computed at line ~333.

## Why it matters

Realistic triggers for `SystemTime` to slip:

- NTP correction at boot or after a long sleep — common on workstations resuming from suspend (and overclockers do reboot/suspend constantly while testing).
- Manual clock change by the user (uncommon but happens).
- VM time drift — common during long autoscans inside a Hyper-V or VMware guest.

When this fires during autoscan, two visible bad things happen:

1. The fluctuation rate spikes from ~0.67 Hz to ~1 Hz silently. The thermal/power-cap detection at `oc_scanner.rs:333-335` uses `thrm_or_pwr_limit_number / in_test_check_number` as a ratio; if `in_test_check_number` accelerates artificially, a *real* throttling event gets diluted and the scanner declares the point stable when the GPU was actually capping.
2. Reproducibility is lost — the same hardware at the same setting can pass or fail an autoscan depending on whether the system clock happened to slip mid-run.

## Reproduction

On Linux (the same applies on Windows but is harder to script):

```bash
sudo date -s "$(date -d '-2 seconds')"   # while autoscan is in the inner loop
```

After that, observe the autoscan emitting `inducing freq fluctuation...` once per second instead of once per 1.5 seconds for the next ~2 seconds of run time.

## Recommended approach

One-line per declaration; no callers change:

```rust
- let mut last_fluctuation = SystemTime::now();
+ let mut last_fluctuation = Instant::now();
```

```rust
- if last_fluctuation.elapsed().unwrap_or(Duration::from_secs(6))
-     >= Duration::from_millis(1500)
+ if last_fluctuation.elapsed() >= Duration::from_millis(1500)
```

```rust
- last_fluctuation = SystemTime::now();
+ last_fluctuation = Instant::now();
```

`Instant::elapsed()` returns `Duration` directly (infallible), so the `.unwrap_or(...)` falls away.

While there, grep `oc_scanner.rs` (and the rest of `auto-optimizer/src/`) for other uses of `SystemTime` in pacing contexts — `oc_scanner.rs` `autoscan_legacy` likely has the same pattern. The only legitimate uses of `SystemTime` in this codebase are for log timestamps; everything else should be `Instant`.

## Severity

Low to medium — the bug is silent and only triggers under clock skew, but when it does it can corrupt the throttling-detection logic that decides whether a point is "stable", which is the whole point of autoscan.

## Out of scope

- Refactoring `apply_fluctuation` or the throttling-detection ratio.
- Adding a configurable fluctuation interval.

Dixiao-L:
## Summary

Fixes #18.

The autoscan inner loop (`pressure_runner::run`) paced its "freq fluctuation" trigger using `SystemTime`, a wall-clock that can move backward. When `elapsed()` returned `Err` after a clock slip, the `unwrap_or(Duration::from_secs(6))` substituted 6 s — larger than the 1.5 s trigger threshold — so the fluctuation step fired on every 1-second loop tick until the clock recovered. This silently inflated `in_test_check_number` and diluted the throttling-detection ratio computed at line 333-335, potentially masking real thermal/power-cap events and declaring an unstable OC point as stable.

`Instant` is monotonic and infallible; its `elapsed()` returns `Duration` directly so the `unwrap_or` chain disappears entirely.

## Changes

- `auto-optimizer/src/oc_scanner.rs`
  - `last_fluctuation` declaration: `SystemTime::now()` → `Instant::now()` (line 229)
  - Elapsed guard: `.elapsed().unwrap_or(Duration::from_secs(6))` → `.elapsed()` (line 246)
  - Reset after firing: `SystemTime::now()` → `Instant::now()` (line 298)
  - Removes now-unused `SystemTime` from the `use std::time` import

No logic changes outside these four lines; all callers are unaffected.

## Test plan

- [ ] `cargo check` passes (verified locally — clean, zero warnings)
- [ ] `cargo clippy` passes
- [ ] Manual: run autoscan, then `sudo date -s "$(date -d '-2 seconds')"` mid-run and confirm `inducing freq fluctuation...` still fires at ~1.5 s intervals rather than every second

https://claude.ai/code/session_01MiHWtF4zRxZhsMEfjTSB86

---
_Generated by [Claude Code](https://claude.ai/code/session_01MiHWtF4zRxZhsMEfjTSB86)_

### DIFF
--- a/auto-optimizer/src/oc_scanner.rs
+++ b/auto-optimizer/src/oc_scanner.rs
@@ -24,7 +24,7 @@ use std::io::Write;
 use std::path::Path;
 use std::process::{Child, Command};
 use std::thread::sleep;
-use std::time::{Duration, Instant, SystemTime};
+use std::time::{Duration, Instant};
 use std::{fs, iter};
 
 
@@ -226,7 +226,7 @@ mod pressure_runner {
                 Ok(mut process) => {
                     let mut exit_code = 1;
                     let test_start_at = Instant::now();
-                    let mut last_fluctuation = SystemTime::now();
+                    let mut last_fluctuation = Instant::now();
                     let mut in_test_check_number = 0;
                     let mut fluctuation_h_l_flag = false;
                     let mut thrm_or_pwr_limit_number = 0;
@@ -243,8 +243,7 @@ mod pressure_runner {
                     sleep(Duration::from_secs(1));
 
                     loop {
-                        if last_fluctuation.elapsed().unwrap_or(Duration::from_secs(6))
-                            >= Duration::from_millis(1500)
+                        if last_fluctuation.elapsed() >= Duration::from_millis(1500)
                         {
                             in_test_check_number += 1;
                             println!("inducing freq fluctuation...");
@@ -295,7 +294,7 @@ mod pressure_runner {
                                 }
                             }
 
-                            last_fluctuation = SystemTime::now();
+                            last_fluctuation = Instant::now();
                         }
 
                         sleep(time::Duration::from_secs(1));


Dixiao-L:
Closing as duplicate. Superseded by #30 (byte-identical diff).

---

PR #27, opened 2026-05-09
fix(auto-optimizer): pre_load_vf_recheck panics on recoverable v-f read failure

### LINKED ISSUE #15: auto-optimizer: `pre_load_vf_recheck` panics whole autoscan on a recoverable v-f read failure
## Summary

`pre_load_vf_recheck` uses `.expect()` to handle the result of `voltage_frequency_check`, panicking the entire autoscan process on the first transient v-f read failure. Every other call site of `voltage_frequency_check` in `oc_scanner.rs` correctly returns or breaks out of the loop on error — this one outlier panics the whole binary instead.

## Evidence

`auto-optimizer/src/oc_scanner.rs:510-514`:

```rust
fn pre_load_vf_recheck(matches: &ArgMatches, point: usize) {
    println!("Waiting for pre-load volt-freq recheck");
    sleep(Duration::from_secs(1));
    voltage_frequency_check(matches.clone(), point).expect("Failed to read v-f info");
}
```

The function is invoked from at least three places inside the autoscan inner loops:

- `oc_scanner.rs:847` — inside the gpuboostv3 short phase per-point ramp
- `oc_scanner.rs:989` — inside the gpuboostv3 long phase
- `oc_scanner.rs:1102` — inside the memory OC phase, immediately before the progress print

Compare with the same function used in the surrounding loop body, `oc_scanner.rs:258-272`:

```rust
match voltage_frequency_check(matches.clone(), cfg.point) {
    Ok(true) => {}
    Ok(false) => {
        println!("Considering trig'd thrm/pwr capping!!!");
        thrm_or_pwr_limit_number += 1;
    }
    Err(e) => {
        eprintln!("Warning: Failed to read v-f info: {}", e);
        force_kill_process(&mut process, "v-f check read failure");
        break;
    }
}
```

That's exactly how the rest of the scanner handles a v-f read failure: log, kill the stress subprocess, break the inner loop, and let the outer loop decide whether to continue or stop. The intent is clearly *not* to abort the whole binary — but `pre_load_vf_recheck` does exactly that.

## Why it matters

`voltage_frequency_check` reads NvAPI VFP state. Transient failures are realistic during autoscan — driver TDR, brief GPU reset, NvAPI getting itself confused after a destabilizing point. When that happens during the 1-second pre-load recheck, the user loses:

- The current scan run (no checkpoint written if the panic fires before the periodic write).
- Any partial results that haven't been flushed to disk.
- The opportunity to recover gracefully (the surrounding loop already knows how to skip a bad point).

For a tool that explicitly carries a "this can crash your driver" warning, the autoscan needs to *expect* driver wobble, not panic on it.

## Reproduction

In a test environment, monkey-patch `voltage_frequency_check` to return `Err(...)` on the second invocation. Run `nvoc set vfp autoscan`. Observe the process panics at `pre_load_vf_recheck` instead of cleanly stopping the current point and proceeding.

## Recommended approach

1. Change the signature to return a result:
   ```rust
   fn pre_load_vf_recheck(matches: &ArgMatches, point: usize) -> Result<(), Error> {
       println!("Waiting for pre-load volt-freq recheck");
       sleep(Duration::from_secs(1));
       voltage_frequency_check(matches.clone(), point)?;
       Ok(())
   }
   ```
2. At the three call sites (`847`, `989`, `1102`), handle the error with the same pattern the surrounding loop uses — log, kill the stress subprocess if one is alive, break/continue.

This is a ~10-line change, mechanical, no behavior change in the happy path.

## Out of scope

- Removing the `matches.clone()` plumbing (covered indirectly by #4 / PR #13 / PR #14).
- Adding retry logic to `voltage_frequency_check` itself.

## Severity

High — single-event silent loss of an entire autoscan run on a *recoverable* error.

Dixiao-L:
## Summary

Fixes #15.

`pre_load_vf_recheck` was calling `voltage_frequency_check(...).expect(...)`, which panics the entire autoscan binary on the first transient NvAPI v-f read failure. Every other call site in `oc_scanner.rs` handles the error gracefully; this was the one outlier.

### Changes

**`auto-optimizer/src/oc_scanner.rs`**

- Changed `pre_load_vf_recheck` signature from `fn(...) -> ()` to `fn(...) -> Result<(), Error>`, replacing `.expect()` with the `?` operator.
- At all three call sites (short phase loop, long phase loop, memory OC phase loop), replaced the bare call with an `if let Err(e)` block that logs a warning and `break`s out of the inner loop — matching the established pattern at line 258.

No behavior change on the happy path. On a transient v-f read failure the scan now breaks out of the current point's inner loop and lets the outer loop decide whether to continue, rather than crashing the process.

## Test plan

- [ ] `cargo check` passes (verified locally — clean compile)
- [ ] `cargo build --release` passes
- [ ] Manual: in a test env, inject a `Err(...)` return from `voltage_frequency_check` during pre-load; confirm the scan exits the inner loop with a warning instead of panicking

https://claude.ai/code/session_01K5CiArGdSHjh6bvBXnW9qA

---
_Generated by [Claude Code](https://claude.ai/code/session_01K5CiArGdSHjh6bvBXnW9qA)_

### DIFF
--- a/auto-optimizer/src/oc_scanner.rs
+++ b/auto-optimizer/src/oc_scanner.rs
@@ -507,10 +507,11 @@ fn apply_short_phase_success_step(
     Some(increase)
 }
 
-fn pre_load_vf_recheck(matches: &ArgMatches, point: usize) {
+fn pre_load_vf_recheck(matches: &ArgMatches, point: usize) -> Result<(), Error> {
     println!("Waiting for pre-load volt-freq recheck");
     sleep(Duration::from_secs(1));
-    voltage_frequency_check(matches.clone(), point).expect("Failed to read v-f info");
+    voltage_frequency_check(matches.clone(), point)?;
+    Ok(())
 }
 
 fn apply_long_phase_failure_step(
@@ -844,7 +845,10 @@ fn run_gpuboostv3_short_phase<V: std::fmt::Display + Copy>(
             Some(*init_core_oc_value - args.common.minimum_delta_core_freq_step),
         )?;
 
-        pre_load_vf_recheck(args.common.matches, point);
+        if let Err(e) = pre_load_vf_recheck(args.common.matches, point) {
+            eprintln!("Warning: Failed to read v-f info: {}", e);
+            break;
+        }
 
         test_num += 1;
         test_code += 1;
@@ -986,7 +990,10 @@ fn run_gpuboostv3_long_phase<V: std::fmt::Display + Copy>(
     writeln!(l, "Initiating Long Test...")?;
 
     loop {
-        pre_load_vf_recheck(args.common.matches, point);
+        if let Err(e) = pre_load_vf_recheck(args.common.matches, point) {
+            eprintln!("Warning: Failed to read v-f info: {}", e);
+            break;
+        }
 
         *test_code += 1;
         log_point_test_header(
@@ -1099,7 +1106,10 @@ fn run_mem_oc_phase<V: std::fmt::Display + Copy>(
         mem_test_num += 1;
         mem_test_code += 1;
 
-        pre_load_vf_recheck(args.common.matches, args.point);
+        if let Err(e) = pre_load_vf_recheck(args.common.matches, args.point) {
+            eprintln!("Warning: Failed to read v-f info: {}", e);
+            break;
+        }
 
         println!(
             "current test progress estimated:{:.2}%",


Dixiao-L:
Closing as duplicate. Same fix area as #28; the warn-and-break approach in #28 was chosen by the majority of competing PRs.

---

PR #36, opened 2026-05-09
fix(validation): replace global-max or-logic with per-element allclose criterion

### LINKED ISSUE #33: cli-stressor-cuda: validation pass criterion lets large per-element errors slip through silently
## Summary

## Evidence

```python
diff = (out_f32 - ref).abs()

```

|---|---|---|
| FP64 | 1e-5 | 1e-5 |
| FP32 | 1e-2 | 1e-2 |
| TF32 | 2e-1 | 2e-1 |
| FP16 | 2e-1 | 2e-1 |
| BF16 | 5e-1 | 5e-1 |
| FP8 E4M3FN | 1.5 | 1.5 |

|---|---|---|
| FP32 | 1e-2 | up to 1.1 |
| FP16 | 2e-1 | up to 22 |
| BF16 | 5e-1 | up to 55 |
| FP8   | 1.5 | up to 165 |

## Why it matters

## Reproduction

## Recommended approach

```
```

Concrete replacement:

```python
diff = (out_f32 - ref).abs()
passed = n_violated == 0
reason = (None if passed else
          f"max_rel={max_rel:.4g}")
```

While there:

## Severity

## Out of scope

Dixiao-L:
## Summary

Fixes #33

Both `cli-stressor-cuda/test.py` and `cli-stressor-opencl/test.py` contained the same broken validation pass criterion:

```python
# BEFORE (buggy)
max_abs = diff.max()
max_rel = max_abs / ref.abs().max()          # global max / global max — different elements!
passed = (max_abs <= atol) or (max_rel <= rtol)  # `or` means either alone can pass
```

This allowed a single badly-corrupted output element to silently pass validation: if the global reference maximum was large (routinely ~110 for N=768 standard-normal matmul), `max_rel` would be small even when `max_abs` was enormous. Concretely, a BF16 element off by 50 units still reported `OK`.

**Root cause**: reducing per-element evidence to global scalars, then combining with `or`, is the exact anti-pattern for detecting localised silent corruption.

## Fix

Both stressors now use the standard `allclose` convention — every element must independently satisfy:

```
|out - ref|  ≤  atol + rtol * |ref|        (per element)
```

```python
# AFTER (fixed)
n_violated = int(violated.sum())
passed = n_violated == 0
reason = f"{n_violated} elements exceed atol+rtol*|ref|; max_abs=..., max_rel=..."
```

The `max_abs` / `max_rel` scalars are still computed and printed in the status line for observability, but the pass/fail decision is made element-wise.

## Scope

- `cli-stressor-cuda/test.py` — torch tensor path  
- `cli-stressor-opencl/test.py` — numpy array path (identical bug, same fix pattern)
- `choose_tolerance()` values are **unchanged** — re-calibration against the stricter criterion is a separate follow-up task (noted in #33)

## Test plan

- [ ] Inject deterministic corruption (`out[0, 0] += 50.0` in a BF16 run) — validation must now report `FAIL` instead of `OK`
- [ ] Clean run (no corruption injected) — validation must continue to report `OK` for all supported precisions
- [ ] Verify the error message format includes the violation count: `N elements exceed atol+rtol*|ref|`

---
_Generated by [Claude Code](https://claude.ai/code/session_01ADnVSh2oX7c611VoQH2VTZ)_

### DIFF
--- a/cli-stressor-cuda/test.py
+++ b/cli-stressor-cuda/test.py
@@ -192,14 +192,21 @@ def validate_precision(
         return False, float("inf"), float("inf"), "validation produced NaN/Inf"
 
     diff = (out_f32 - ref).abs()
-    max_abs = float(diff.max().item())
-    ref_abs = float(ref.abs().max().item())
-    max_rel = max_abs / (ref_abs + 1e-12)
     abs_thr, rel_thr = choose_tolerance(spec.name)
 
-    # 只要绝对误差或相对误差其中一个仍在可接受范围内，就不判失败。
-    passed = (max_abs <= abs_thr) or (max_rel <= rel_thr)
-    reason = None if passed else f"error too large: abs={max_abs:.4g}, rel={max_rel:.4g}"
+    # Per-element bound: atol + rtol * |ref|  (same convention as torch.allclose / numpy.allclose).
+    # Using a global max-of-diff vs max-of-ref with an `or` would let a single badly-corrupted
+    # element pass whenever the global reference max is large enough to make max_rel look small.
+    elementwise_bound = abs_thr + rel_thr * ref.abs()
+    violated = diff > elementwise_bound
+    n_violated = int(violated.sum().item())
+    max_abs = float(diff.max().item())
+    ref_abs_max = float(ref.abs().max().item())
+    max_rel = max_abs / (ref_abs_max + 1e-12)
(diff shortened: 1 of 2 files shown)


Dixiao-L:
Closing as duplicate. Per-element validation fix is fully included in #40 (which also adds compute-only TFLOPS).

---

Issue #38, opened 2026-05-09
cli-stressor-opencl: same TFLOPS measurement bias as #34 (burst flops over total wall time)


Dixiao-L:
## Summary

Same defect class as #34 (CUDA stressor), in the OpenCL stressor. `result.tflops` divides burst-only flops by total wall time, so the headline number depends on validation cadence, warmup count, allocator timing, and `print` volume — not on sustained GPU throughput. The `inst_tflops` printed each window is computed correctly; the cumulative summary number is biased low.

## Evidence

`cli-stressor-opencl/test.py:653-685`:

```python
op_start = time.monotonic()
executed_iters = burst_iters
for _ in range(burst_iters):
    enqueue_gemm(runtime, bundle, active_buffers, size, transpose_a, transpose_b)
runtime.queue.finish()
op_elapsed = time.monotonic() - op_start

if target_burst_s > 0 and op_elapsed < target_burst_s:
    multiplier = max(1, int(math.ceil(target_burst_s / max(op_elapsed, 1e-9))))
    extra_iters = burst_iters * (multiplier - 1)
    if extra_iters > 0:
        extra_start = time.monotonic()
        for _ in range(extra_iters):
            enqueue_gemm(runtime, bundle, active_buffers, size, transpose_a, transpose_b)
        runtime.queue.finish()
        op_elapsed += time.monotonic() - extra_start
        executed_iters += extra_iters

inst_tflops = (2 * (size**3) * executed_iters / op_elapsed) / 1e12       # ← correct
elapsed_total = time.monotonic() - start

print(...)

result.iterations += executed_iters
result.total_flops += int(2 * (size**3) * executed_iters)                # ← burst-only
result.elapsed_s = time.monotonic() - start                              # ← total wall
if result.elapsed_s > 0:
    result.tflops = (result.total_flops / result.elapsed_s) / 1e12       # ← mixed
```

Same shape as #34: numerator counts only burst flops (warmup excluded by design). Denominator includes everything between iterations — buffer reallocation when size changes (lines 635-641), `enqueue_copy` refresh of inputs (lines 645-646), the validation sidecar (lines 688-712), the prints, the warmup loop on the next iteration (lines 649-651). Reported sustained TFLOPS depends on operational settings, not on GPU throughput.

[...]

Issue #37, opened 2026-05-09
cli-stressor-opencl: same broken validation pass criterion as #33 — large per-element errors slip through


Dixiao-L:
## Summary

Same defect class as #33 (CUDA stressor), in the OpenCL stressor. The validation pass criterion uses tensor-wide max-of-each then `or` between abs and rel tolerance. This admits per-element errors large enough to indicate severe silent corruption whenever the maximum reference value is comparable to or larger than the tolerance. Since detecting silent corruption is the *primary purpose* of the OpenCL stressor's sidecar validation, this defeats the gate.

The two stressors should be fixed together (or with a shared helper) so the criterion stays consistent.

## Evidence

`cli-stressor-opencl/test.py:553-565`:

```python
out_f32 = out.astype(np.float32, copy=False)
if not np.isfinite(out_f32).all():
    return False, float("inf"), float("inf"), "validation produced NaN/Inf"

diff = np.abs(out_f32 - ref)
max_abs = float(np.max(diff))
ref_abs = float(np.max(np.abs(ref)))
max_rel = max_abs / (ref_abs + 1e-12)
abs_thr, rel_thr = choose_tolerance(spec.name)

passed = (max_abs <= abs_thr) or (max_rel <= rel_thr)
reason = None if passed else f"error too large: abs={max_abs:.4g}, rel={max_rel:.4g}"
return passed, max_abs, max_rel, reason
```

Tolerance table (lines 226-233):

| precision | abs_thr | rel_thr |
|---|---|---|
| FP64 | 1e-5 | 1e-5 |
| FP32 | 1e-2 | 1e-2 |
| FP16 | 2e-1 | 2e-1 |

Same numerical analysis as #33: with `--validate-size 768` and N(0,1) inputs, `max(|ref|) ≈ 110`, so the relative branch implicitly admits per-element absolute errors up to `rel_thr · 110`. For FP16 that's ~22; for FP32 it's ~1.1. The `or` combinator means satisfying either bound is enough to pass.

The `if not np.isfinite(out_f32).all()` check on line 554 catches NaN/Inf, which is good — but a corrupted output that's merely *large and wrong* (rather than non-finite) sails through.

## Why it matters

Same as #33 — the validation sidecar is the only stability gate inside the stressor; if the gate biases toward pass, the stressor returns "OK" on most realistic silent-corruption signatures and the autoscan caller treats unstable points as stable.

[...]

PR #40, opened 2026-05-09
fix(stressor): per-element validation criterion and compute-only TFLOPS (closes #33, #34, #37, #38)

### LINKED ISSUE #33: cli-stressor-cuda: validation pass criterion lets large per-element errors slip through silently

### LINKED ISSUE #34: cli-stressor-cuda: cumulative TFLOPS divides burst-only flops by total wall time — systematically low and setting-dependent
## Summary

The CUDA stressor's reported "TFLOPS" in the per-precision summary divides burst-only flop counts by total wall time, where total wall time includes warmup, sidecar validation, deletes, allocator reset, and `print` calls. The instantaneous `inst_tflops` printed each window is correct; the cumulative `result.tflops` is systematically biased low and depends on operational settings (`--validate-interval`, `--warmup-iters`, print volume) that have nothing to do with sustained GPU throughput.

This matters because the OpenCL stressor's `database.md` uses TFLOPS as a baseline-comparison signal for stability regressions. If autoscan or any downstream consumer compares the cumulative TFLOPS to that baseline, the comparison is unreliable across changes in validation cadence or warmup count.

## Evidence

`cli-stressor-cuda/test.py:271-294`:

```python
# 正式压力段。
op_start = time.monotonic()
for _ in range(burst_iters):
    aa = a.t() if transpose_a else a
    bb = b.t() if transpose_b else b
    _ = torch.mm(aa, bb)
synchronize_device(device)
op_elapsed = time.monotonic() - op_start
inst_tflops = (2 * (size ** 3) * burst_iters / op_elapsed) / 1e12       # ← correct
elapsed_total = time.monotonic() - start

print(...)

result.iterations += burst_iters
result.total_flops += int(2 * (size**3) * burst_iters)                  # ← burst-only
result.elapsed_s = time.monotonic() - start                              # ← total wall
if result.elapsed_s > 0:
    result.tflops = (result.total_flops / result.elapsed_s) / 1e12       # ← mixed
```

`result.elapsed_s` is updated again at line 334 after the `while True` exits, but it's still the same total-wall metric.

What `result.elapsed_s` includes between two burst windows:

- The `del a, b; empty_device_cache(device)` allocator dance (line 296-297).
- The optional sidecar validation, which runs a 768×768 FP32→FP64 round-trip on CPU (lines 300-323) — this is significant on slow CPUs.
- The `print(...)` calls (line 282-288, 309-312).
- The next-iteration matrix construction (`make_random_matrix` × 2, line 261-262) — these allocate fresh tensors and copy CPU→GPU.
- The warmup iteration loop (lines 264-269) for the next window.

Numerator (`total_flops`) excludes warmup. Denominator (`elapsed_s`) includes everything. Result is biased.

Quick numerical example. Default settings: `--burst-iters 6`, `--warmup-iters 3`, `--validate-interval 10`, sizes biased toward 127-1023. A typical inner loop:

- 3 warmup iters at ~512² ≈ 0.5 ms each → 1.5 ms
- 6 burst iters at ~512² ≈ 1 ms each → 6 ms
- alloc/del/empty_cache → 5-30 ms (PyTorch CUDA caching allocator)
- validate every ~10 windows → 768³·2·1 / (CPU FP64) ≈ 100-500 ms

Effective ratio for cumulative TFLOPS: numerator counts the 6 ms of burst flops, denominator counts ~12-50 ms of wall time. Reported TFLOPS is 1/4 to 1/8 of true sustained burst throughput.

Compare to `inst_tflops` (line 279), which divides burst flops by burst-only `op_elapsed`. That one is correct and matches what users expect when they say "TFLOPS".

## Why it matters

1. The `cli-stressor-opencl/database.md` table was contributed by community members benchmarking real hardware. If they ran the stressor with `--validate-interval 0` (every iteration), they got one number; if they ran with the default 10 s, they got a different number on the same hardware. The baseline is non-reproducible across operational settings.
2. The `print_summary` line (`test.py:355-362`) prominently shows `tflops` as a peer of `iters` and `time`, with no annotation that it includes non-compute time. A user comparing two GPU runs, one with longer validation, will see a "regression" that's purely an artifact of the meter.
3. If autoscan ever starts using this number to gate "is the GPU running healthy", spurious regressions or improvements will be attributed to the GPU.

## Reproduction

```
$ python test.py --duration 60 --validate-interval 1   --precisions fp16
... reports TFLOPS X
$ python test.py --duration 60 --validate-interval 999 --precisions fp16
... reports TFLOPS Y, where Y > X consistently on the same hardware
```

The two runs are doing the same computation; only the validation cadence differs. The reported sustained TFLOPS will differ noticeably.

## Recommended approach

Track compute time separately:

```python
# accumulate burst time only
result.compute_s = 0.0  # add to dataclass

# inside the loop, after op_elapsed is computed:
result.total_flops += int(2 * (size**3) * burst_iters)
result.compute_s += op_elapsed

# at the end:
if result.compute_s > 0:
    result.tflops = (result.total_flops / result.compute_s) / 1e12
result.elapsed_s = time.monotonic() - start  # keep total wall for display
```

In `print_summary`, print both:

```
fp16   OK | iters=N | wall=Ts | compute=Cs | sustained=X TFLOPS | wallclock_eff=Y%
```

Where `Y = compute_s / wall_s · 100` lets the reader see what fraction of the run was actually GPU compute. This also makes validation cost visible.

Mirror the same fix in `cli-stressor-opencl/test.py` (open a follow-up if it has the same bug) and *then* re-baseline `database.md` so the table's "TFLOPS" column has a consistent definition.

## Severity

Medium. Not a correctness bug for the GPU itself, but a measurement bug that undermines the stressor's role as a benchmark/regression detector.

## Out of scope

- Changing tolerances or workload (separate issue tracks the validation correctness — see the validation pass-criterion issue).
- Re-baselining `database.md` (follow-up after the fix).

Dixiao-L:
## Summary

Fixes four bugs across `cli-stressor-cuda/test.py` and `cli-stressor-opencl/test.py`.

---

### Fixes #33 and #37 — Broken validation pass criterion

**Root cause**: both stressors reduced per-element diff to a single `max_abs` scalar, divided it by the global `max(|ref|)` scalar (from a potentially *different* element), then combined with `or` so satisfying *either* loose bound was enough to pass. With N=768 standard-normal matmul, `ref.abs().max()` is routinely ~110, making the relative branch trivially pass even for BF16 elements off by 50 units.

**Fix**: replace with the standard `allclose` convention — every element must independently satisfy:

```
|out - ref| ≤ atol + rtol * |ref|        (per element)
```

A failing run now reports the violation count (`N elements exceed atol+rtol*|ref|`), plus `max_abs` and `max_rel` for observability.

**Note**: `choose_tolerance()` values are unchanged — re-calibration against the stricter criterion is a separate follow-up (noted in #33/#37).

---

### Fixes #34 and #38 — TFLOPS measurement bias

**Root cause**: both stressors computed `result.tflops` by dividing burst-only flop counts by total wall time (`elapsed_s`), which includes warmup iterations, sidecar validation, allocator resets (`empty_cache`/buffer reallocation), `print` calls, and host↔device copies. This made the headline TFLOPS systematically low and dependent on `--validate-interval`, `--warmup-iters`, and `--input-refresh-interval` — settings unrelated to GPU throughput — making the `database.md` baseline non-reproducible across operational configurations.

**Fix**: add `compute_s: float` to `StressResult` and accumulate only `op_elapsed` (burst-kernel wall time, after `synchronize`). `result.tflops` is computed from `compute_s` only. `elapsed_s` retains the full wall time for display.

`print_summary` now shows:
```
FP16   OK | iters=... | wall= 90.0s | compute= 12.3s | 245.67 TFLOPS | eff=13.7% | ...
```

The `eff=` column makes validation/warmup overhead immediately visible.

---

## Scope

- `cli-stressor-cuda/test.py` — validation fix + TFLOPS fix
- `cli-stressor-opencl/test.py` — validation fix + TFLOPS fix
- `choose_tolerance()` values — **unchanged**
- `sys.exit(1)` on validation failure — **unchanged** (confirmed intentional: immediate abort prevents driver lock-up on Linux with no TDR)

## Test plan

- [ ] Inject `out[0, 0] += 50.0` in a BF16 CUDA run — validation must now report `FAIL` with `1 elements exceed atol+rtol*|ref|`
- [ ] Clean CUDA run — validation must continue to report `OK` for all precisions
- [ ] Run CUDA stressor with `--validate-interval 1` vs `--validate-interval 999` on the same hardware — reported TFLOPS must now be consistent
- [ ] Same corruption injection and clean run for OpenCL stressor
- [ ] Run OpenCL with `--validate-interval 1 --input-refresh-interval 1` vs `--validate-interval 999 --input-refresh-interval 64` — TFLOPS must now be consistent

https://claude.ai/code/session_01X93XY3NZKnFYepVaQRKrRP

---
_Generated by [Claude Code](https://claude.ai/code/session_01X93XY3NZKnFYepVaQRKrRP)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,6 @@
 .codex
+__pycache__/
+*.pyc
+*.pyo
+.venv/
+*.egg-info/
--- a/cli-stressor-cuda/test.py
+++ b/cli-stressor-cuda/test.py
@@ -22,6 +22,7 @@ class StressResult:
     iterations: int = 0
     total_flops: int = 0
     elapsed_s: float = 0.0
+    compute_s: float = 0.0
     tflops: float = 0.0
     validations: int = 0
     validation_failures: int = 0
@@ -192,14 +193,21 @@ def validate_precision(
         return False, float("inf"), float("inf"), "validation produced NaN/Inf"
 
     diff = (out_f32 - ref).abs()
-    max_abs = float(diff.max().item())
-    ref_abs = float(ref.abs().max().item())
-    max_rel = max_abs / (ref_abs + 1e-12)
     abs_thr, rel_thr = choose_tolerance(spec.name)
 
-    # 只要绝对误差或相对误差其中一个仍在可接受范围内，就不判失败。
-    passed = (max_abs <= abs_thr) or (max_rel <= rel_thr)
-    reason = None if passed else f"error too large: abs={max_abs:.4g}, rel={max_rel:.4g}"
+    # Per-element bound: atol + rtol * |ref|  (same convention as torch.allclose).
+    # A global max-of-diff vs max-of-ref with `or` lets a single badly-corrupted element
+    # pass whenever the global reference max is large enough to make max_rel look small.
+    elementwise_bound = abs_thr + rel_thr * ref.abs()
+    violated = diff > elementwise_bound
+    n_violated = int(violated.sum().item())
+    max_abs = float(diff.max().item())
+    ref_abs_max = float(ref.abs().max().item())
+    max_rel = max_abs / (ref_abs_max + 1e-12)
+    passed = n_violated == 0
+    reason = (None if passed else
+              f"{n_violated} elements exceed atol+rtol*|ref|; max_abs={max_abs:.4g}, "
+              f"max_rel={max_rel:.4g}")
     return passed, max_abs, max_rel, reason
 
 
@@ -289,9 +297,7 @@ def run_stress_for_precision(
 
             result.iterations += burst_iters
             result.total_flops += int(2 * (size**3) * burst_iters)
-            result.elapsed_s = time.monotonic() - start
-            if result.elapsed_s > 0:
-                result.tflops = (result.total_flops / result.elapsed_s) / 1e12
+            result.compute_s += op_elapsed
 
             del a, b
             empty_device_cache(device)
@@ -332,8 +338,8 @@ def run_stress_for_precision(
             time.sleep(0)
 
     result.elapsed_s = time.monotonic() - start
-    if result.elapsed_s > 0:
-        result.tflops = (result.total_flops / result.elapsed_s) / 1e12
+    if result.compute_s > 0:
+        result.tflops = (result.total_flops / result.compute_s) / 1e12
     return result
 
 
@@ -352,11 +358,15 @@ def print_summary(device_name: str, total_memory_gb, results):
             status = "OK" if (r.first_error is None and r.validation_failures == 0) else "FAIL"
         if status != "OK":
             overall_ok = False
+        wallclock_eff = (f"{100 * r.compute_s / r.elapsed_s:.1f}%"
+                         if r.elapsed_s > 0 else "n/a")
         print(
             f"{r.precision:12} {status:4} | "
             f"iters={r.iterations:8d} | "
-            f"time={r.elapsed_s:7.1f}s | "
+            f"wall={r.elapsed_s:7.1f}s | "
+            f"compute={r.compute_s:6.1f}s | "
             f"{r.tflops:8.2f} TFLOPS | "
+            f"eff={wallclock_eff:6} | "
             f"validations={r.validations:3d} | "
             f"val_fail={r.validation_failures:3d} | "
             f"max_abs={r.max_abs_error:.3g} | max_rel={r.max_rel_error:.3g}"
--- a/cli-stressor-opencl/test.py
+++ b/cli-stressor-opencl/test.py
@@ -97,6 +97,7 @@ class StressResult:
     iterations: int = 0
     total_flops: int = 0
     elapsed_s: float = 0.0
+    compute_s: float = 0.0
     tflops: float = 0.0
     validations: int = 0
     validation_failures: int = 0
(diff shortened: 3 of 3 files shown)


Dixiao-L:
Superseded by #59, which consolidates the stressor validation/TFLOPS fixes (#33, #34, #37, #38) on the current `main`. Closing as part of PR-queue triage.

---

Issue #35, opened 2026-05-09
cli-stressor-cuda: `sys.exit(1)` from inside inner loop kills multi-precision sweep before summary


Dixiao-L:
## Summary

`run_stress_for_precision` calls `sys.exit(1)` from inside the inner stress loop on the first validation failure or runtime exception. The exit fires *inside* `run_stress_for_precision`, so the function never returns its accumulated `StressResult`, the `for spec in precisions` loop in `main()` never iterates to the next precision, and `print_summary` never runs. The carefully-designed cumulative counters (`validation_failures`, `max_abs_error`, `max_rel_error`) are written to the result struct one line before exit, then thrown away.

This is inconsistent with the dataclass design and with the multi-precision sweep architecture.

## Evidence

`cli-stressor-cuda/test.py:316-328`:

```python
if not passed:
    result.validation_failures += 1
    if result.first_error is None:
        result.first_error = reason
        result.first_error_at_s = time.monotonic() - start
        sys.exit(1)                          # ← exits the whole process
next_validate = time.monotonic() + max(0.0, validate_interval_s)
validation_seed += 1

except Exception as exc:
    result.first_error = f"runtime error: {exc}"
    result.first_error_at_s = time.monotonic() - start
    sys.exit(1)                              # ← same pattern
```

The dataclass declares (`test.py:18-31`):

```python
@dataclass
class StressResult:
    ...
    validations: int = 0
    validation_failures: int = 0       # plural — designed to count multiple failures
    max_abs_error: float = 0.0
    max_rel_error: float = 0.0
    first_error: Optional[str] = None
    first_error_at_s: Optional[float] = None
```

The `print_summary` formatter prints them as peers (`test.py:355-362`):

```python
print(
    f"{r.precision:12} {status:4} | "
    f"iters={r.iterations:8d} | "
    f"time={r.elapsed_s:7.1f}s | "
    f"{r.tflops:8.2f} TFLOPS | "
    f"validations={r.validations:3d} | "
    f"val_fail={r.validation_failures:3d} | "
    f"max_abs={r.max_abs_error:.3g} | max_rel={r.max_rel_error:.3g}"
)
```

`val_fail={r.validation_failures:3d}` exists in the format string for a reason — the design clearly intends to accumulate and report multiple failures. The `sys.exit(1)` on first failure makes the field unreachable.

[...]

PR #41, opened 2026-05-09
fix(stressors): fix TFLOPS measurement bias and sys.exit in inner loop (closes #34, #35, #38)

### LINKED ISSUE #34: cli-stressor-cuda: cumulative TFLOPS divides burst-only flops by total wall time — systematically low and setting-dependent

Dixiao-L:
## Summary

Fixes three bugs in `cli-stressor-cuda/test.py` and `cli-stressor-opencl/test.py`.

---

### Issue #34 / #38 — cumulative TFLOPS measurement bias (CUDA + OpenCL)

**Root cause:** `result.tflops` divided burst-only flop counts by total wall time. The denominator included warmup iterations, validation sidecars, buffer allocation/refresh (`create_device_buffers`, `enqueue_copy`), and `print` calls — none of which are GPU compute. The instantaneous `inst_tflops` printed each window was already correct; only the cumulative summary was wrong. Result depended on `--validate-interval`, `--warmup-iters`, `--input-refresh-interval`, and allocator timing — not on GPU throughput.

**Fix:** Add `compute_s: float = 0.0` to `StressResult` in both stressors. Accumulate `result.compute_s += op_elapsed` after each burst (including the min-burst adaptive extension in OpenCL). Use `compute_s` for TFLOPS everywhere. Keep `elapsed_s` as wall clock for display. `print_summary` now shows:

```
FP16          OK | iters=     N | wall=  90.0s | compute=  42.3s | eff= 47.0% |    8.45 TFLOPS | ...
```

The `eff` field (compute_s / elapsed_s × 100%) makes overhead visible so users can reason about why cumulative TFLOPS differs from instantaneous TFLOPS.

---

### Issue #35 — `sys.exit(1)` from inside CUDA inner loop kills multi-precision sweep

**Root cause:** `run_stress_for_precision` called `sys.exit(1)` on the first validation failure and on any runtime exception. This killed the entire process before the `for spec in precisions` loop in `main()` could iterate to the next precision, and before `print_summary` ran. The `validation_failures` counter, `max_abs_error`, and `max_rel_error` fields were populated one line before the exit and then discarded. The OpenCL stressor already used `break` in equivalent paths — this was a CUDA-only inconsistency.

**Fix (Option A — fail soft, report at end, matching the dataclass design):**
- **Validation failure:** remove `sys.exit(1)`. The loop continues accumulating data. `first_error` is recorded on the first failure; subsequent failures increment `validation_failures`. All remaining precisions still run.
- **Runtime exception:** replace `sys.exit(1)` with `break` so the while loop exits cleanly and the function returns its partial `StressResult`.
- **`print_summary`:** remove `sys.exit(1)` from inside it. `main()` already exits non-zero via `if not overall_passed: sys.exit(1)` — the internal exit was dead code after any failure and prevented the closing `"=" * 72` separator from printing.

---

## Test plan

- [x] `python3 -m py_compile` — both files, no syntax errors
- [x] `ruff check` — all checks pass
- [x] 14 unit tests in `cli-stressor-cuda/test_fixes.py` and `cli-stressor-opencl/test_fixes.py` — all pass without a GPU
  - `TestComputeSField` — verifies `compute_s` exists, accumulates correctly, drives TFLOPS, returns 0 when no compute time, gives identical TFLOPS for same work across different validate intervals
  - `TestNoSysExitInInnerLoop` — verifies validation failures don't exit, all precisions accumulate results, exceptions don't exit
  - `TestPrintSummaryNoExit` — verifies `print_summary` does not call `sys.exit`
- [ ] Live GPU run: verify `inst_tflops` (per-window) ≈ summary TFLOPS after fix
- [ ] Live GPU run: inject `out[0, 0] += 50.0` in validation; confirm FP16 reports FAIL but BF16 and FP32 still run and produce summary output

## Out of scope

- Validation pass criterion (#33 / #37) — addressed by PR #36
- Re-baselining `cli-stressor-opencl/database.md` — should be done after this merges, since the TFLOPS definition changes

---
_Generated by [Claude Code](https://claude.ai/code/session_01UuZNcpefWh8ei7kCgG1pM6)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 .codex
+__pycache__/
+*.pyc
+*.pyo
--- a/cli-stressor-cuda/test.py
+++ b/cli-stressor-cuda/test.py
@@ -22,6 +22,7 @@ class StressResult:
     iterations: int = 0
     total_flops: int = 0
     elapsed_s: float = 0.0
+    compute_s: float = 0.0  # burst-only GPU compute time (excludes warmup, validation, alloc)
     tflops: float = 0.0
     validations: int = 0
     validation_failures: int = 0
@@ -289,9 +290,10 @@ def run_stress_for_precision(
 
             result.iterations += burst_iters
             result.total_flops += int(2 * (size**3) * burst_iters)
+            result.compute_s += op_elapsed
             result.elapsed_s = time.monotonic() - start
-            if result.elapsed_s > 0:
-                result.tflops = (result.total_flops / result.elapsed_s) / 1e12
+            if result.compute_s > 0:
+                result.tflops = (result.total_flops / result.compute_s) / 1e12
 
             del a, b
             empty_device_cache(device)
@@ -318,22 +320,21 @@ def run_stress_for_precision(
                     if result.first_error is None:
                         result.first_error = reason
                         result.first_error_at_s = time.monotonic() - start
-                        sys.exit(1)
                 next_validate = time.monotonic() + max(0.0, validate_interval_s)
                 validation_seed += 1
 
         except Exception as exc:
             result.first_error = f"runtime error: {exc}"
             result.first_error_at_s = time.monotonic() - start
-            sys.exit(1)
+            break
 
         # 若 burst 很短，避免 CPU 忙等。
         if op_elapsed < 0.01:
             time.sleep(0)
 
     result.elapsed_s = time.monotonic() - start
-    if result.elapsed_s > 0:
-        result.tflops = (result.total_flops / result.elapsed_s) / 1e12
+    if result.compute_s > 0:
+        result.tflops = (result.total_flops / result.compute_s) / 1e12
     return result
 
 
@@ -352,10 +353,13 @@ def print_summary(device_name: str, total_memory_gb, results):
             status = "OK" if (r.first_error is None and r.validation_failures == 0) else "FAIL"
         if status != "OK":
             overall_ok = False
+        eff = f"{100 * r.compute_s / r.elapsed_s:.1f}%" if r.elapsed_s > 0 else "n/a"
         print(
             f"{r.precision:12} {status:4} | "
             f"iters={r.iterations:8d} | "
-            f"time={r.elapsed_s:7.1f}s | "
+            f"wall={r.elapsed_s:7.1f}s | "
+            f"compute={r.compute_s:6.1f}s | "
+            f"eff={eff:>6} | "
             f"{r.tflops:8.2f} TFLOPS | "
             f"validations={r.validations:3d} | "
             f"val_fail={r.validation_failures:3d} | "
@@ -372,7 +376,6 @@ def print_summary(device_name: str, total_memory_gb, results):
         print("- 在当前测试窗口内，没有出现明显计算错误或验证失败。")
     else:
         print("- 至少有一个精度模式出现错误、验证失败或不支持。")
-        sys.exit(1)
     print("=" * 72)
 
 
@@ -521,7 +524,7 @@ def main():
             if res.supported:
                 overall_passed = False
         else:
-            print(f"  结果: completed without detected error")
+            print("  结果: completed without detected error")
         print(f"  累计: {res.iterations} 次 matmul, {res.tflops:.2f} TFLOPS")
 
     print_summary(device_name, total_memory_gb, results)
--- a/cli-stressor-cuda/test_fixes.py
+++ b/cli-stressor-cuda/test_fixes.py
@@ -1,0 +1,212 @@
+"""
+Unit tests for fixes to cli-stressor-cuda/test.py:
+  - Issue #35: sys.exit(1) removed from inner loop; multi-precision sweep completes
+  - Issue #34: compute_s tracks burst-only time; TFLOPS uses compute_s not elapsed_s
+"""
+
+import sys
+import types
+import unittest
+from dataclasses import dataclass
+from typing import Optional
+from unittest.mock import MagicMock, patch
+
+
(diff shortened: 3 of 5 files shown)


Dixiao-L:
Superseded by #59, which consolidates the stressor TFLOPS / `sys.exit` fixes (#34, #35, #38) on the current `main`. Closing as part of PR-queue triage.

---

Issue #47, opened 2026-05-09
auto-optimizer: `select_gpus` silently drops invalid `--gpu` entries in mixed-validity lists


Dixiao-L:
## Summary

`select_gpus` and `select_gpu_ids` (in `basic_func.rs`) silently drop invalid GPU index entries from a multi-value `--gpu` argument. If the user runs `nvoc status --gpu 0,5` on a 2-GPU system, the function returns `Ok([gpu_0])` with no error, no warning, and no indication that GPU 5 didn't exist. The user thinks both GPUs were processed; only one was.

The error path only fires when *every* requested GPU is invalid (the `selected.is_empty()` check at the end). Mixed valid/invalid input is silently filtered.

## Evidence

`auto-optimizer/src/basic_func.rs:252-299`:

```rust
pub fn select_gpus<'a>(
    gpus: &'a [Gpu],
    gpu: Option<clap::parser::ValuesRef<'_, String>>,
) -> anyhow::Result<Vec<&'a Gpu>, Error> {
    let selected = match gpu {
        Some(values) => {
            let inputs = values
                .map(|s| parse_gpu_id(s.as_str()))
                .collect::<anyhow::Result<Vec<_>, _>>()
                .map_err(|e| Error::Custom(e.to_string()))?;

            let mut result = Vec::new();
            for input in inputs {
                // ① 人类可读序号（强语义，禁止 fallback）
                if input < 256 {
                    if let Some(g) = gpus.get(input) {
                        result.push(g);
                        continue;
                    } else {
                        // index 不存在，直接认为无效
                        continue;             // ← silently skip!
                    }
                }

                // ② 直接 GPU ID（dec / hex）
                if let Some(g) = gpus.iter().find(|g| g.id() == input) {
                    result.push(g);
                    continue;
                }

                // ③ legacy fallback（只允许 input >= 256）
                let legacy = input << 8;
                if let Some(g) = gpus.iter().find(|g| g.id() == legacy) {
                    result.push(g);
                    continue;
                }
                // No match in any of the three lookup forms — silently drop
            }
            result
        }
        None => gpus.iter().collect(),
    };

    if selected.is_empty() {
        Err(Error::DeviceNotFound)
    } else {
        Ok(selected)
    }
}
```

`select_gpu_ids` (lines 330-373) has the same shape and the same silent-drop behavior.

[...]

Issue #46, opened 2026-05-09
auto-optimizer: `TestResolution::R640x384.downgrade()` returns itself — autoscan resolution loop hangs


Dixiao-L:
## Summary

`TestResolution::downgrade()` returns *itself* for both `R640x384` and `R400x300`. The `R640x384` case is clearly an unfinished edit — there's a commented-out line right next to it showing the intended `R576x360` target. The `R400x300` case is the lowest resolution and self-loop is plausibly intentional, but it's undocumented and the function returns `Option<Self>` (which suggests the lowest-rung sentinel was meant to be `None`, not `Some(self)`). Either way, callers that use `downgrade` in a loop will spin forever on these terminal states.

The autoscan resolution-picker is exactly such a caller: when a stress run is unstable, it walks down the resolution ladder by calling `downgrade` repeatedly until stability is reached.

## Evidence

`auto-optimizer/src/basic_func.rs:103-127`:

```rust
impl TestResolution {
    pub fn downgrade(self) -> Option<Self> {
        match self {
            TestResolution::R3600x1920 => Some(TestResolution::R3072x1600),
            TestResolution::R3072x1600 => Some(TestResolution::R2560x1440),
            ...
            TestResolution::R720x480  => Some(TestResolution::R640x384),
            TestResolution::R640x384  => Some(TestResolution::R640x384),    // ← self-loop
            // TestResolution::R640x384 => Some(TestResolution::R576x360),  // ← commented-out intended target
            TestResolution::R576x360  => Some(TestResolution::R400x300),
            TestResolution::R400x300  => Some(TestResolution::R400x300),    // ← self-loop, terminal state
            // Lowest resolution, no downgrade available
        }
    }
}
```

Two distinct issues:

1. **`R640x384 => Some(R640x384)`** — clearly a bug. The commented-out neighbor proves the intended target was `R576x360`. This was probably meant to be uncommented at some point and wasn't. The function does have `R576x360 => Some(R400x300)` two lines later, so `R576x360` is a valid step in the ladder; the gap is just `R640x384 → R576x360`.
2. **`R400x300 => Some(R400x300)`** — terminal state. The trailing comment "Lowest resolution, no downgrade available" suggests the author meant to return `None` here. The function signature `Option<Self>` supports it. Returning `Some(self)` is a contract violation against the doc comment "Get the next lower resolution".

## Why it matters

A caller pattern like:

```rust
let mut res = current_resolution;
loop {
[...]

Issue #45, opened 2026-05-09
auto-optimizer: `get_primary_screen_size_raw` uses fixed 220-byte buffer for `DEVMODEW` — stack-corruption risk on extended driver layouts


Dixiao-L:
## Summary

`get_primary_screen_size_raw` calls Win32 `EnumDisplaySettingsW` with a fixed-size `[u8; 220]` stack buffer and a hand-coded byte-offset map for `DEVMODEW`. Two related risks:

1. The 220-byte size assumption is hard-coded with a comment that says it's the standard layout, but `DEVMODEW` is documented as extensible — printer drivers and newer Windows versions can extend the structure. If `EnumDisplaySettingsW` is ever called against a device whose driver returns a larger `DEVMODEW`, the kernel writes past the 220-byte buffer and corrupts the calling thread's stack.
2. `[u8; 220]` is 1-byte aligned in Rust. `DEVMODEW` requires 4-byte (DWORD) alignment for several of its fields. The kernel's writes through this buffer assume the proper alignment of `DEVMODEW` start. On x86/x64 unaligned access happens to work, but the unsafe contract is not honored as documented.

This is `auto-optimizer`'s only direct Win32 FFI for display info, so the blast radius is limited to `set_resolution` and the autoscan's resolution-picker, but the failure mode is the worst class — silent stack corruption.

## Evidence

`auto-optimizer/src/basic_func.rs:28-76`:

```rust
#[cfg(windows)]
fn get_primary_screen_size_raw() -> (u32, u32) {
    use std::ffi::c_int;
    unsafe extern "system" {
        fn EnumDisplaySettingsW(
            lpsz_device_name: *const u16,
            i_mode_num: u32,
            lp_dev_mode: *mut u8,                 // ← raw byte pointer
        ) -> c_int;
    }
    // DEVMODEW（wingdi.h）Win32 ABI 固定布局，x86-64 下：
    //   +  0  dmDeviceName[32]  = 64 字节
    //   ...
    //   +172  dmPelsWidth       = 4 字节
    //   +176  dmPelsHeight      = 4 字节
    //   …（后续字段省略，总结构体大小 220 字节）
    const DEVMODEW_SIZE: usize = 220;
    const OFFSET_DM_SIZE: usize = 68;
    const OFFSET_PELS_WIDTH: usize = 172;
    const OFFSET_PELS_HEIGHT: usize = 176;
[...]

Issue #44, opened 2026-05-09
auto-optimizer: `update_csv_with_load_and_margin` non-atomic temp-file rename + hardcoded `./ws/` path can lose user's VFP CSV


Dixiao-L:
## Summary

`update_csv_with_load_and_margin` writes its output to the hardcoded path `./ws/temp.csv`, then deletes the user's input file, then renames the temp file into place. The sequence is non-atomic and the path is hardcoded relative to the *current working directory*. Two failure modes:

1. If the process crashes (panic, SIGKILL, power loss) between the `remove_file` and the `rename`, the user's original CSV is gone forever and the new content is in `./ws/temp.csv`.
2. If `./ws/` doesn't exist relative to CWD when the function runs (a common case — the user runs `nvoc` from anywhere, not necessarily from the repo root), `File::create("./ws/temp.csv")` errors out.

Both are real risks during the autoscan/fix_result workflow that this function supports.

## Evidence

`auto-optimizer/src/oc_profile_function.rs:217-271`:

```rust
fn update_csv_with_load_and_margin(
    file_path: &str,
    default_frequencies: Vec<u32>,
    default_frequencies_load: Vec<u32>,
    minimum_delta_core_freq_step: i32,
    legacy_flag: bool,
) -> Result<(), Error> {
    let mut rdr = ReaderBuilder::new()
        .has_headers(true)
        .from_path(file_path)?;
    let mut wtr = WriterBuilder::new()
        .has_headers(true)
        .from_writer(File::create("./ws/temp.csv")?);          // ← hardcoded relative path

    ...
    wtr.flush()?;

    // Replace original file with updated file
    fs::remove_file(file_path)?;                                // ← delete original
    fs::rename("./ws/temp.csv", file_path)?;                    // ← rename temp into place

    Ok(())
}
```

Three problems in three lines:

- Hardcoded `./ws/temp.csv` — depends on `./ws/` existing relative to the process's CWD. Frontends that invoke `auto-optimizer` from a fixed working dir might create it; users running from a terminal won't.
- `remove_file` then `rename` is a non-atomic replace. `fs::rename` over an existing file is itself atomic on Linux and Windows. The explicit `remove_file` before rename is *unnecessary* and creates the gap.
- The temp file is in a *different directory* (`./ws/`) from the original, so the `rename` may cross filesystems and fall back to copy+delete, which is also non-atomic.

## Why it matters

[...]

PR #52, opened 2026-05-09
fix(auto-optimizer): panics, silent GPU drops, unsafe FFI, atomic writes (closes #42–#47)

### LINKED ISSUE #42: auto-optimizer: `fix_result` and friends panic on plausible bad CSV inputs (assert_eq!, unwrap, usize underflow)
## Summary

`fix_result` (the V-F curve post-processor invoked by `nvoc set vfp fix_result`) and the helpers it calls (`interpolate_deltas`, `get_key_points_indices`, `key_point_extractor`, `extract_default_frequencies`) panic on a wide range of plausible bad-CSV inputs: a CSV with fewer than 4 non-zero key points, a row missing a column, a CSV with no rows whose voltage exceeds 680 mV, a row whose voltage column isn't a valid u32. The panics use `assert_eq!`, `.unwrap()`, and `usize` underflow on `records.len() - 1`. None of them are reachable through clean Rust error propagation.

The `fix_result` workflow is exactly where users hand-edit a CSV — between an autoscan run and the import — so brittle parsing here is the most likely place to lose user work.

### A. `assert_eq!` panics on a CSV with < 4 key points

`oc_profile_function.rs:508-538`:

```rust
fn get_key_points_indices(lines: &[Vec<String>]) -> (usize, usize, usize, usize) {
    let mut key_indices = Vec::new();
    for (i, columns) in lines.iter().enumerate() {
        let freq = columns.get(1).and_then(|s| s.parse::<u32>().ok()).unwrap_or(0);
        let delta = columns.get(2).and_then(|s| s.parse::<i32>().ok()).unwrap_or(0);
        if freq != 0 && delta != 0 {
            key_indices.push(i);
            if key_indices.len() == 4 { break; }
        }
    }
    assert_eq!(key_indices.len(), 4, "Need exactly 4 key points in ultrafast mode");
    (key_indices[0], key_indices[1], key_indices[2], key_indices[3])
}
```

Called from `interpolate_deltas` at line 542. Called from `fix_result` at line 626 when `cfg.is_ultrafast` is true. A CSV the user produced with the autoscan in non-ultrafast mode (which often has fewer non-zero deltas), then post-processed with `fix_result --ultrafast`, panics here. The user gets an opaque `Need exactly 4 key points in ultrafast mode` and no recovery path.

### B. `.unwrap()` chain on column parses inside `interpolate_deltas`

`oc_profile_function.rs:549-563`:

```rust
let p1_v = lines[p1_idx][0].parse::<u32>().unwrap();
let p2_v = lines[p2_idx][0].parse::<u32>().unwrap();
let p3_v = lines[p3_idx][0].parse::<u32>().unwrap();
let p4_v = lines[p4_idx][0].parse::<u32>().unwrap();

if maxq_flag {
    p1_d = lines[p1_idx][2].parse::<i32>().unwrap() - minimum_delta_step;
    p2_d = lines[p2_idx][2].parse::<i32>().unwrap() - minimum_delta_step;
    p3_d = lines[p3_idx][2].parse::<i32>().unwrap() - 2 * minimum_delta_step;
    p4_d = lines[p4_idx][2].parse::<i32>().unwrap() - 2 * minimum_delta_step;
} else {
    ...
}
```

`get_key_points_indices` validated `columns[1]` and `columns[2]` are parseable. It did NOT validate `columns[0]` (voltage). It also indexes with `[0]` and `[2]` directly — `lines[p1_idx][0]` panics if the row is empty. Same in the next loop body at line 567: `let current_v = lines[i][0].parse::<u32>().unwrap();` — every row's voltage column must parse, including rows that get_key_points_indices already classified as "non-key" because freq or delta failed to parse.

A row like `,,5000` (empty voltage column) → panic.

### C. `key_point_extractor` panics on `records.len() - 1` when no rows match the voltage filter

`oc_profile_function.rs:946-993`:

```rust
let mut records: Vec<(usize, u32, u32, i32)> = Vec::new();

for (idx, result) in rdr.records().enumerate() {
    let record = result?;
    ...
    if voltage > 680000 {
        records.push((idx, default_freq, default_freq_load, margin_bin));
    }
    ...
}

for i in 0..records.len() - 1 {       // ← underflow if records is empty
    ...
}
```

If no row has `voltage > 680000` (a plausible scenario for a low-voltage / Max-Q profile), `records.len() == 0`, and `records.len() - 1` underflows `usize` to `usize::MAX`. Debug build: panics with `attempt to subtract with overflow`. Release build: loops effectively forever (until the indexing panic on `records[i]`).

### D. `extract_default_frequencies` direct-indexes the CSV row

`oc_profile_function.rs:192-208`:

```rust
for result in rdr.records() {
    let record = result?;
    let default_frequency_load: u32;
    if legacy_flag {
        default_frequency_load = record[1].parse()?;     // panic if record has < 2 fields
    } else {
        default_frequency_load = record[3].parse()?;     // panic if record has < 4 fields
    }
    default_frequencies_load.push(default_frequency_load);
}
```

`csv::StringRecord::index` panics on out-of-range; the `parse()?` only handles parse errors after the row has been indexed safely.

### E. `assert_eq!` in log parsing

`oc_profile_function.rs:898`:

```rust
if let Some(point) = extract_value(line, "point: #") {
    assert_eq!(last_voltage_point.unwrap(), point);  // panic on log inconsistency
}
```

A truncated or hand-edited log → panic. This is in `export_vfp_from_log`, called from `nvoc set vfp export_log`.

## Why it matters

The `fix_result` and `export_vfp_from_log` paths are exactly where a user is most likely to feed in a hand-touched file — they sit between an autoscan run (which produced the CSV/log) and the import that writes the final VFP curve to the GPU. Panics here destroy intermediate work without a clean error message, and at least one path (C) silently corrupts memory in release builds.

`fix_result` also writes its output by overwriting the input file (line 689: `File::create(&cfg.output)`). A panic mid-write leaves the user with a truncated CSV — see also the related issue about non-atomic temp-file rename.

## Reproduction

```
# A. < 4 key points
$ printf 'voltage,frequency,delta\n800000,2000000,0\n850000,2050000,0\n900000,2100000,0\n' > scan.csv
$ nvoc set vfp fix_result --ultrafast --vfpath scan.csv --output fixed.csv
# panics: "Need exactly 4 key points in ultrafast mode"

# C. No row above 680 mV
$ printf 'voltage,...\n600000,...\n650000,...\n' > scan.csv
$ nvoc set vfp fix_result --vfpath scan.csv --output fixed.csv
# panics: "attempt to subtract with overflow"  (debug build)

# D. Short row
$ printf 'voltage\n800000\n' > scan.csv
$ nvoc set vfp fix_result --vfpath scan.csv --output fixed.csv
# panics: index out of bounds: the len is 1 but the index is 1
```

## Recommended approach

Treat the CSV as untrusted input throughout. A small set of helpers + propagated `Result<(), Error>` covers all five sites:

```rust
fn parse_col<T: std::str::FromStr>(row: &[String], idx: usize, what: &str) -> Result<T, Error>
where T::Err: std::fmt::Display {
    row.get(idx)
        .ok_or_else(|| Error::from(format!("CSV row missing column {} ({})", idx, what)))?
        .parse::<T>()
        .map_err(|e| Error::from(format!("CSV column {} ({}) not parseable: {}", idx, what, e)))
}
```

Then:

- A. Change `get_key_points_indices` to return `Result<(usize,usize,usize,usize), Error>` and produce a clear error like `"ultrafast mode needs ≥ 4 key points (rows with non-zero freq AND delta), found N"`.
- B. Replace each `lines[p1_idx][0].parse::<u32>().unwrap()` with `parse_col::<u32>(&lines[p1_idx], 0, "voltage")?`.
- C. Guard with `if records.is_empty() { return Err(Error::from("no rows above 680 mV")); }` before the `records.len() - 1` loop.
- D. Replace `record[1].parse()?` / `record[3].parse()?` with helper-based access.
- E. Replace `assert_eq!(last_voltage_point.unwrap(), point)` with a logged warning + `continue` (the log line is informational, not authoritative).

While there: also fix `export_single_point`'s `Box::leak(sum.to_string().into_boxed_str())` at line 134 — same `Box::leak` pattern as #10, applied per CSV row in a loop. A single export of a 100-row CSV leaks 100 strings.

## Severity

High (within `auto-optimizer`). Each path is reachable by the documented user workflow (`fix_result`, `export_log`, `import`) and produces a panic the user can't recover from cleanly. The release-build silent corruption in path C is the worst.

## Out of scope

- The `Box::leak` in `export_single_point` (mention as duplicate of #10's pattern).
- Restructuring the V-F curve schema or moving away from CSV.
- The interpolation correctness (separate, see the `linear_interpolate` divide-by-zero issue).

Dixiao-L:
## Summary

Fixes six bugs in `auto-optimizer/src/basic_func.rs` and `auto-optimizer/src/oc_profile_function.rs`.

---

### Issue #46 — `TestResolution::R640x384.downgrade()` self-loop hangs autoscan

**Root cause:** `R640x384 => Some(R640x384)` — clearly an unfinished edit (a commented-out `R576x360` target sits right next to it). `R400x300 => Some(R400x300)` violated the `Option` contract (should be `None` for the lowest rung). Any autoscan resolution loop calling `downgrade()` in a loop would spin forever once it hit `R640x384`.

**Fix:** `R640x384 => Some(R576x360)`, `R400x300 => None`. Two-line change.

---

### Issue #47 — `select_gpus` / `select_gpu_ids` silently drop invalid `--gpu` entries

**Root cause:** Both functions silently `continue`d past any unmatched GPU index. `nvoc status --gpu 0,99` on a 2-GPU system would return only GPU 0 with no warning; the user believed both GPUs were acted on.

**Fix:** Replace the silent skip with `Error::Custom("no GPU matches --gpu N; use nvoc list to see available indices")` in both `select_gpus` and `select_gpu_ids`. Any invalid index is now a hard error (Option A from the issue).

---

### Issue #44 — `update_csv_with_load_and_margin` non-atomic CSV replace loses user data

**Root cause:** Temp file written to hardcoded `./ws/temp.csv` (fails outside repo root), then `remove_file` + `rename` — a crash in the 50 ms gap between them deletes the user's hours-long autoscan CSV permanently. Cross-filesystem rename also fell back to non-atomic copy+delete.

**Fix:** Write temp file as `.filename.tmp-<pid>` next to the destination (same filesystem). Drop the explicit `remove_file` — `fs::rename` over an existing path is already atomic on POSIX. No more hardcoded `./ws/` dependency.

---

### Issue #43 — `linear_interpolate` divide-by-zero and `u32` underflow write wrong deltas to GPU

**Root cause:** `(v2-v1)` could be zero (flat V-F region), yielding `NaN` silently cast to `0` — clobbering OC deltas with "no overclock" with no warning. `(current_v-v1)` was unsigned subtraction and underflowed to ~`u32::MAX` when `current_v < v1` (non-monotonic hand-edited CSV), producing absurdly large deltas in release builds.

**Fix:** Return `Result<i32, Error>`. Guard `v1 == v2` (return average of endpoints). Use ordered endpoints so subtraction is always non-negative; return a descriptive error if `current_v` is out of range.

---

### Issue #42 — Multiple `panic!` / `.unwrap()` / `usize` underflow sites on bad CSV inputs

**Root cause:** Five panic sites in the `fix_result` / `export_log` workflow — exactly where users supply hand-edited files:

| Site | Mechanism | Effect |
|------|-----------|--------|
| `get_key_points_indices` | `assert_eq!(len, 4)` | Opaque panic on <4 key points |
| `interpolate_deltas` | `.unwrap()` on every column parse | Panic on missing/non-numeric column |
| `key_point_extractor` | `records.len() - 1` when `records` is empty | usize underflow → silent memory corruption in release builds |
| `extract_default_frequencies` | `record[N]` direct index | Panic on short row |
| `export_vfp_from_log` | `assert_eq!(last_voltage_point.unwrap(), point)` | Panic on truncated/hand-edited log |

**Fix:**
- Add `parse_col<T>(row, idx, what)` helper — safe `get` + `parse` with descriptive error messages.
- `get_key_points_indices` → `Result<(usize,usize,usize,usize), Error>` with a clear message.
- `interpolate_deltas` → `Result<(), Error>`; use `parse_col` throughout; propagate `?`.
- `key_point_extractor`: guard `records.is_empty()` before the subtraction loop.
- `extract_default_frequencies`: `record.get(idx)` with explicit OOB error.
- `export_vfp_from_log`: replace `assert_eq!` with `eprintln!` warning + `continue`.

---

### Issue #45 — `get_primary_screen_size_raw` uses 1-byte-aligned `[u8; 220]` for `DEVMODEW`

**Root cause:** `DEVMODEW` DWORD fields require 4-byte alignment. The raw `[u8; 220]` buffer is 1-byte aligned, violating the unsafe contract. On x86/x64 this happens to work but trips sanitizers and would be a real issue on ARM64.

**Fix:** Wrap the buffer in `#[repr(C, align(4))] struct AlignedBuf([u8; 256])` — the compiler ensures 4-byte alignment; 256 bytes provides headroom over the 220-byte standard layout without changing the offset map or the `dmSize` value advertised to the kernel.

---

## Test plan

- [x] `cargo check` (dev profile) — passes, zero warnings
- [x] `cargo check --release` — passes, zero warnings
- [ ] Live Windows: run `nvoc status --gpu 0,99` on a 1-GPU system; confirm hard error rather than silent partial result
- [ ] Live Windows: run autoscan at R640x384 stability floor; confirm loop steps down to R576x360 then R400x300 then stops
- [ ] Live Windows: run `fix_result --ultrafast` against a CSV with only 3 key points; confirm clean error message
- [ ] Live Windows: run `fix_result --ultrafast` against a CSV with equal consecutive voltages; confirm non-zero deltas in output

---

_Generated by [Claude Code](https://claude.ai/code/session_019mbJsmrNiCpFy822PgEyUx)_

### DIFF
--- a/auto-optimizer/src/basic_func.rs
+++ b/auto-optimizer/src/basic_func.rs
@@ -47,24 +47,34 @@ fn get_primary_screen_size_raw() -> (u32, u32) {
     //   +176  dmPelsHeight      = 4 字节  ← OFFSET_PELS_HEIGHT
     //   …（后续字段省略，总结构体大小 220 字节）
     // 参考：https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-devmodew
-    const DEVMODEW_SIZE: usize = 220;
+    //
+    // The buffer is over-allocated to 256 bytes and given `align(4)` so that the
+    // DWORD fields the kernel writes (dmPelsWidth, dmPelsHeight, …) land at their
+    // natural 4-byte alignment.  The standard DEVMODEW size (220) is still written
+    // into dmSize so the kernel knows the exact buffer capacity we advertise.
+    const DEVMODEW_STD_SIZE: usize = 220;
     const OFFSET_DM_SIZE: usize = 68;
     const OFFSET_PELS_WIDTH: usize = 172;
     const OFFSET_PELS_HEIGHT: usize = 176;
     const ENUM_CURRENT_SETTINGS: u32 = 0xFFFF_FFFF;
+
+    #[repr(C, align(4))]
+    struct AlignedBuf([u8; 256]);
+
     unsafe {
-        let mut buf = [0u8; DEVMODEW_SIZE];
-        buf[OFFSET_DM_SIZE] = DEVMODEW_SIZE as u8;
-        buf[OFFSET_DM_SIZE + 1] = (DEVMODEW_SIZE >> 8) as u8;
-        let ret = EnumDisplaySettingsW(std::ptr::null(), ENUM_CURRENT_SETTINGS, buf.as_mut_ptr());
+        let mut buf = AlignedBuf([0u8; 256]);
+        buf.0[OFFSET_DM_SIZE] = DEVMODEW_STD_SIZE as u8;
+        buf.0[OFFSET_DM_SIZE + 1] = (DEVMODEW_STD_SIZE >> 8) as u8;
+        let ret =
+            EnumDisplaySettingsW(std::ptr::null(), ENUM_CURRENT_SETTINGS, buf.0.as_mut_ptr());
         if ret != 0 {
             let w = u32::from_le_bytes(
-                buf[OFFSET_PELS_WIDTH..OFFSET_PELS_WIDTH + 4]
+                buf.0[OFFSET_PELS_WIDTH..OFFSET_PELS_WIDTH + 4]
                     .try_into()
                     .unwrap(),
             );
             let h = u32::from_le_bytes(
-                buf[OFFSET_PELS_HEIGHT..OFFSET_PELS_HEIGHT + 4]
+                buf.0[OFFSET_PELS_HEIGHT..OFFSET_PELS_HEIGHT + 4]
                     .try_into()
                     .unwrap(),
             );
@@ -118,11 +128,9 @@ impl TestResolution {
             TestResolution::R800x600 => Some(TestResolution::R768x576),
             TestResolution::R768x576 => Some(TestResolution::R720x480),
             TestResolution::R720x480 => Some(TestResolution::R640x384),
-            TestResolution::R640x384 => Some(TestResolution::R640x384),
-            // TestResolution::R640x384 => Some(TestResolution::R576x360),
+            TestResolution::R640x384 => Some(TestResolution::R576x360),
             TestResolution::R576x360 => Some(TestResolution::R400x300),
-            TestResolution::R400x300 => Some(TestResolution::R400x300),
-            // Lowest resolution, no downgrade available
+            TestResolution::R400x300 => None,
         }
     }
 
@@ -262,28 +270,20 @@ pub fn select_gpus<'a>(
 
             let mut result = Vec::new();
             for input in inputs {
-                // ① 人类可读序号（强语义，禁止 fallback）
-                if input < 256 {
-                    if let Some(g) = gpus.get(input) {
-                        result.push(g);
-                        continue;
-                    } else {
-                        // index 不存在，直接认为无效
-                        continue;
-                    }
-                }
-
-                // ② 直接 GPU ID（dec / hex）
-                if let Some(g) = gpus.iter().find(|g| g.id() == input) {
-                    result.push(g);
-                    continue;
-                }
-
-                // ③ legacy fallback（只允许 input >= 256）
-                let legacy = input << 8;
-                if let Some(g) = gpus.iter().find(|g| g.id() == legacy) {
-                    result.push(g);
-                    continue;
+                let matched = if input < 256 {
+                    // ① 人类可读序号（强语义，禁止 fallback）
+                    gpus.get(input)
+                } else {
+                    // ② 直接 GPU ID（dec / hex），③ legacy fallback（只允许 input >= 256）
+                    gpus.iter().find(|g| g.id() == input)
+                        .or_else(|| gpus.iter().find(|g| g.id() == input << 8))
+                };
+                match matched {
+                    Some(g) => result.push(g),
+                    None => return Err(Error::Custom(format!(
+                        "no GPU matches --gpu {}; use `nvoc list` to see available indices",
+                        input
+                    ))),
                 }
             }
             result
@@ -340,24 +340,20 @@ pub fn select_gpu_ids(
 
             let mut result = Vec::new();
             for input in inputs {
-                if input < 256 {
-                    if let Some(id) = gpu_ids.get(input) {
-                        result.push(*id);
-                        continue;
-                    } else {
-                        continue;
-                    }
-                }
-
-                if let Some(&id) = gpu_ids.iter().find(|&&id| id as usize == input) {
-                    result.push(id);
-                    continue;
-                }
-
-                let legacy = (input as u32) << 8;
-                if let Some(&id) = gpu_ids.iter().find(|&&id| id == legacy) {
-                    result.push(id);
-                    continue;
+                let matched = if input < 256 {
+                    gpu_ids.get(input).copied()
+                } else {
+                    gpu_ids.iter().find(|&&id| id as usize == input).copied().or_else(|| {
+                        let legacy = (input as u32) << 8;
+                        gpu_ids.iter().find(|&&id| id == legacy).copied()
+                    })
+                };
+                match matched {
+                    Some(id) => result.push(id),
+                    None => return Err(Error::Custom(format!(
+                        "no GPU matches --gpu {}; use `nvoc list` to see available indices",
+                        input
+                    ))),
                 }
             }
             result
--- a/auto-optimizer/src/oc_profile_function.rs
+++ b/auto-optimizer/src/oc_profile_function.rs
@@ -189,20 +189,22 @@ fn extract_default_frequencies(file_path: &str, legacy_flag: bool) -> Result<Vec
         .from_path(file_path)?;
     let mut default_frequencies_load = Vec::new();
 
-    for result in rdr.records() {
+    for (row_idx, result) in rdr.records().enumerate() {
         let record = result?;
-        let default_frequency_load: u32;
-
-        if legacy_flag {
-            default_frequency_load = record[1].parse()?;
-        }
-        // Read only frequency column
-        else {
-            default_frequency_load = record[3].parse()?;
-        }
-        // Read only default_frequency column
-
-        default_frequencies_load.push(default_frequency_load);
+        let col_idx = if legacy_flag { 1 } else { 3 };
+        let val = record.get(col_idx).ok_or_else(|| {
+            Error::from(format!(
+                "CSV row {} missing column {} (default_frequency_load)",
+                row_idx, col_idx
+            ))
+        })?;
+        let freq = val.parse::<u32>().map_err(|e| {
+            Error::from(format!(
+                "CSV row {} column {} (default_frequency_load) not parseable as u32: {}",
+                row_idx, col_idx, e
+            ))
+        })?;
+        default_frequencies_load.push(freq);
     }
     Ok(default_frequencies_load)
 }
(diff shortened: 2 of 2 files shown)


Dixiao-L:
Reviewed the diff against issues #42–#47. Each is correctly addressed. Summary:

| Issue | Fix verdict |
|---|---|
| **#42** (CSV panic chain) | ✅ Correct. `assert_eq!` → typed `Error` return; `parse_col` helper introduced and applied across `interpolate_deltas` and `extract_default_frequencies`; `records.len() - 1` underflow guarded with `is_empty()` check; `assert_eq!(last_voltage_point.unwrap(), point)` replaced with a logged warning + continue. All five sites I called out are fixed. |
| **#43** (`linear_interpolate` div-by-zero / underflow) | ✅ Exactly the recommended shape. `v1 == v2` → average of endpoints (no silent 0); ordering normalized; explicit range check; signature now `Result<i32, Error>` with `?` propagation up through `interpolate_deltas` → `fix_result`. |
| **#44** (non-atomic CSV rename + hardcoded `./ws/`) | ✅ Sibling temp file in destination dir, `process::id()` for uniqueness, single `fs::rename(&tmp, dest)`. Hardcoded `./ws/` gone. |
| **#45** (DEVMODEW size / alignment) | ✅ Defensible. Uses `#[repr(C, align(4))] struct AlignedBuf([u8; 256])` — over-allocates by 36 bytes and aligns to 4. The kernel still respects `dmSize = 220`, so it won't write more than that, but the alignment fix removes the 1-byte-aligned hazard. Doesn't switch to the typed `windows::DEVMODEW` (which I'd suggested as the cleaner option), but the in-place fix is sufficient for the immediate safety concern. Fine as-is. |
| **#46** (`R640x384.downgrade()` self-loop) | ✅ Both fixes applied: `R640x384 => Some(R576x360)` and `R400x300 => None`. Exactly what the issue requested. |
| **#47** (silent GPU drop) | ✅ Strict-error variant adopted. `Err(Error::Custom(format!("no GPU matches --gpu {}; use \`nvoc list\` to see available indices", input)))` — friendly message, points users at the discovery command. Applied symmetrically to `select_gpus` and `select_gpu_ids`. |

## Smaller observations / nits (not blocking)

- **#43 `linear_interpolate` flat-region semantics.** The fix uses `(d1 + d2) / 2` for `v1 == v2`. That's what the issue suggested, but worth a comment explaining the choice — a future reader might wonder why averaging vs. picking either endpoint. Two lines of doc on the function would close the loop.
- **#43 `(d1 + d2) / 2` rounds toward zero.** For `d1 = 5, d2 = -8`, `(5 + -8) / 2 = -1` (rust integer division truncates toward zero). For OC deltas this is fine; just noting.
- **#42 the `current_vs` pre-collection** in `interpolate_deltas` (line ~620) is a clean way to avoid the borrow-checker dance. Good.
- **#45** if you ever want to fully eliminate the size assumption, switching to `windows::Win32::Graphics::Gdi::DEVMODEW` with `mem::size_of::<DEVMODEW>()` for `dmSize` would be the canonical version. Pulling in the `windows` crate is a non-trivial dep, so understandable to defer.

## Conflict assessment with other open PRs

- **PR #14** (NVML rewrite) doesn't touch `basic_func.rs::select_gpus` or `oc_profile_function.rs`, so no conflict with this PR. Conflict will be in `main.rs` (this PR doesn't modify it; #14 does).
- **PR #39** (`physicsdolphin`'s big refactor) modifies `basic_func.rs` (+22/-7) and `oc_profile_function.rs` (+51/-28). The `basic_func.rs` patches in #39 are small and likely won't conflict with this PR's `select_gpus` rewrite. The `oc_profile_function.rs` overlap is larger — recommend rebasing this PR if #39 lands first, or vice versa.

## Recommendation

This is a high-quality fix-up PR. All six issues are correctly addressed with the recommended approaches. **Mergeable as-is once the maintainer signs off** (it's a draft PR, but the work is complete from my read). The minor nits above are polish, not blockers.

Nice work — this is exactly the right shape for an "address several review issues at once" PR.

---

Dixiao-L:
This PR is superseded by #56, which covers all the same issues (#42–#47) plus #15, #17, #18, #48, #50, #51, #53 in a single consolidated fix. Suggest closing this one once #56 is reviewed.

---

Dixiao-L:
Superseded by #57, which covers the same fixes (#42–#47) along with 8 additional root-cause fixes. Suggest closing in favour of #57.

---

Dixiao-L:
Superseded by #57 — strict subset (#42–#47 are all included in the consolidated 14-bug PR). Closing as part of PR-queue triage.

---

Issue #49, opened 2026-05-09
auto-optimizer: arg_help.rs lacks range validators on numeric CLI args — upstream source of multiple write-path overflow bugs


Dixiao-L:
## Summary

`auto-optimizer/src/arg_help.rs` defines ~80 numeric CLI arguments (delta_khz, voltage_uv, MHz, point indices, intervals, limits) and almost none of them have range validators. Inputs flow as `String` from clap into handlers, where they are parsed with `.parse::<u32>()?` / `.parse::<i32>()?` and then fed directly to NvAPI / NVML write functions. The result is the recurring class of bug seen in #16 (set_legacy_clocks_nvapi MHz overflow), the linear_interpolate underflows in #43, and the NVML power-limit overflow filed alongside this issue.

A single round of validator additions at the clap layer would block this entire class of bugs upstream.

## Evidence

A grep for clap range/value-parser calls in `arg_help.rs` (851 lines):

```
$ grep -nE 'value_parser|range\(|RangedI64|RangedU64' arg_help.rs
42:                .value_parser(OutputFormat::possible_values().to_vec())
75-128: POSSIBLE_BOOL.to_vec()  (boolean enums × 6)
158, 431, 462: value_parser(["1", "2", "all"])  (cooler index)
167, 176, 185, 261: enum possible_values()
668: value_parser(|v: &str| { v.parse::<i32>().map_err(|_| ...) })
807, 845: value_parser(["aggressive", "traditional"])
```

That's it. Every other numeric input — including:

- `--delta` for VFP point delta (kHz, can go to ±500,000+)
- `--voltage` for voltage offsets (μV, range varies by GPU but typically ±100,000)
- `--core` and `--memory` for `set legacy-clock` (MHz; #16's overflow path)
- `--power-limit` for NVML (W; the companion NVML overflow issue's path)
- `--point` for VFP point index (0..~120)
- `--gpu` for GPU index (0..N)
- `--temperature` for NVML temperature threshold (°C)
- All the autoscan tuning args (`point_start`, `interval_s`, `wait_time_sec`, `freq_step_exp`, etc.)

— is parsed as a free-form `String`, with no `value_parser(value_parser!(T).range(...))` and no custom validator. The handler functions then `.parse::<T>()?` and pass the result onward without range checks.

The one custom validator at line 668-671 (for `minus_bin`) only checks "can be parsed as i32"; it doesn't bound the magnitude.

## Why it matters

[...]

Issue #48, opened 2026-05-09
auto-optimizer: NVML power-limit and memory-clock setters overflow on user input — same shape as #16, no upstream clamp


Dixiao-L:
## Summary

`set_nvml_power_limit` and `set_nvml_memory_clock_offset` perform unchecked arithmetic on user-supplied values before calling NVML. Same defect class as #16 (which covered `set_legacy_clocks_nvapi`), in the NVML write path. There's no upstream range validator on the CLI side, no clamp at the boundary, and the multiplication can overflow `u32`/`i32` silently (release) or panic (debug).

NVML will likely reject absurd values, but the failure mode is "kernel rejects with an opaque error" rather than "CLI tells the user 'that's not a sane GPU power'".

## Evidence

`auto-optimizer/src/oc_get_set_function_nvml.rs:226-249`:

```rust
pub fn set_nvml_power_limit(gpu_id: u32, limit_w: u32) -> Result<(), Error> {
    let nvml = match Nvml::init() { ... };
    let pci_bus_num = gpu_id / 256;
    let device_count = match nvml.device_count() { ... };
    for i in 0..device_count {
        if let Ok(mut dev) = nvml.device_by_index(i) {
            if let Ok(pci_info) = dev.pci_info() {
                if pci_info.bus == pci_bus_num {
                    let limit_mw = limit_w * 1000;                          // ← overflow path
                    match dev.set_power_management_limit(limit_mw) {
                        ...
                    }
                }
            }
        }
    }
    ...
}
```

For `limit_w > u32::MAX / 1000 ≈ 4_294_967`, `limit_w * 1000` overflows `u32`. In debug this panics; in release it wraps silently and writes a tiny value into `set_power_management_limit`.

Same shape at `oc_get_set_function_nvml.rs:214`:

```rust
match dev.set_clock_offset(
    nvml_wrapper::enum_wrappers::device::Clock::Memory,
    pstate,
    (offset * 2) as i32,                                                    // ← overflow path
) {
```

For `offset` passed as `i32`, `offset * 2` overflows when `|offset| > i32::MAX / 2 ≈ 1.07B`. The `as i32` cast doesn't help — the multiplication overflows before the cast.

Looking at `arg_help.rs`, none of these numeric args have a `value_parser(value_parser!(u32).range(...))` or any other range constraint. The CLI accepts any `u32` / `i32` and passes it straight through.

## Why it matters

Same as #16 but on the NVML path:

[...]

Issue #43, opened 2026-05-09
auto-optimizer: `linear_interpolate` divides by zero / `u32`-underflows on plausible V-F CSV shapes — silent wrong deltas


Dixiao-L:
## Summary

`linear_interpolate` divides by `(v2 - v1)` without guarding against `v1 == v2`, and subtracts `(current_v - v1)` as `u32 - u32` without guarding against `current_v < v1`. Both produce silent wrong answers in realistic edge cases of the V-F curve interpolation: the first yields NaN that gets cast to 0 (under Rust's float-to-int saturation rules), the second underflows `u32` to a huge number that then divides into a meaningless ratio. Neither failure is reported.

The function is the core of `interpolate_deltas`, which is the core of `fix_result --ultrafast` — silently producing wrong deltas here writes wrong values into the V-F curve that's later applied to the GPU.

## Evidence

`oc_profile_function.rs:496-506`:

```rust
fn linear_interpolate(v1: u32, d1: i32, v2: u32, d2: i32, current_v: u32, delta_step: i32) -> i32 {
    let ratio = (current_v - v1) as f64 / (v2 - v1) as f64;
    let interpolated = (d2 as f64 - d1 as f64) * ratio + d1 as f64;

    let rounded = if interpolated >= delta_step as f64 {
        (interpolated / delta_step as f64).floor() * delta_step as f64
    } else {
        0.0
    };
    rounded as i32
}
```

Two failure modes:

### A. `v2 == v1` → division by zero → NaN → silent 0

If two consecutive key points share a voltage (a flat region of the V-F curve where the user / scanner wrote identical voltages but different frequencies), `v2 - v1 == 0`. The cast to `f64` first means `0.0_f64 / 0.0_f64 == NaN`. Then `interpolated == NaN`, the `interpolated >= delta_step` check is false (NaN comparisons return false), and the function returns `0.0 as i32 == 0`.

The caller (`interpolate_deltas`, line 575/579/581) writes `0` as the delta for that row — which is "no overclock" — into the CSV. The user's careful intermediate CSV gets clobbered with zeros around the flat region with no warning.

### B. `current_v < v1` → `u32` underflow

`(current_v - v1) as f64` evaluates `u32 - u32` first, then casts. If `current_v < v1`, in release the subtraction wraps to a near-`u32::MAX` value (~4.29e9), which then casts cleanly to `f64` and produces a huge `ratio`. The interpolated value is then absurdly large or absurdly small.

[...]

PR #54, opened 2026-05-09
auto-optimizer: fix 7 correctness/safety/UX bugs (#17/#42/#43/#44/#46/#47/#48/#49)

### LINKED ISSUE #17: auto-optimizer: memory autoscan progress percent is integer-divided, always prints 0.00% / 1.00%

Dixiao-L:
## Summary

Root-cause fixes for 7 open issues across `auto-optimizer`. Each fix targets the upstream source of the bug rather than patching symptoms.

---

### #46 — `TestResolution::downgrade()` infinite loops (critical correctness)

**Root cause**: Two self-referential match arms:
- `R640x384 => Some(R640x384)` — jumps to itself forever
- `R400x300 => Some(R400x300)` — lowest resolution loops instead of terminating

**Fix** (`basic_func.rs`):
```rust
// Before
TestResolution::R640x384 => Some(TestResolution::R640x384),
TestResolution::R400x300 => Some(TestResolution::R400x300),

// After
TestResolution::R640x384 => Some(TestResolution::R576x360),
TestResolution::R400x300 => None,
```

Unit tests added: `test_resolution_downgrade_ladder`, `test_resolution_r640x384_not_self_loop`, `test_resolution_r400x300_is_terminal`.

---

### #43 — `linear_interpolate` NaN and u32 underflow (data corruption)

**Root cause** (`oc_profile_function.rs`):
1. When `v1 == v2` (flat V-F region): `(current_v - v1) as f64 / (v2 - v1) as f64` → `0.0/0.0 = NaN`. All comparisons with NaN return false, so the result silently becomes 0.
2. When `current_v < v1`: `current_v - v1` underflows as `u32`, wrapping to ~4.29 billion → garbage interpolated delta.

**Fix**: Return `Result<i32, Error>`. Special-case the flat region (`d1/2 + d2/2`). Normalise to `(lo_v, hi_v)` and bounds-check `current_v` before subtraction.

---

### #42 — Panics and silent data corruption in ultrafast/MaxQ profile (multiple)

**Root causes** (`oc_profile_function.rs`):

- **A** `get_key_points_indices`: panicked with index-out-of-bounds when fewer than 4 key rows exist → now returns `Result` with a descriptive error.
- **B** `interpolate_deltas`: scattered `.unwrap()` on CSV column parses → replaced with `parse_col::<T>()` helper that propagates `Result`.
- **C** `key_point_extractor`: `records[0..records.len()-1]` panics on empty slice → guarded with an early `is_empty()` check.
- **D** `extract_default_frequencies`: `record[col]` index operator on `StringRecord` panics on short rows → replaced with `.get(col).ok_or_else(...)`.
- **E** `export_vfp_from_log`: `assert_eq!` on log point ordering panics on out-of-order logs → replaced with a `continue` + `eprintln!` warning.

---

### #44 — Non-atomic CSV write in `update_csv_with_load_and_margin` (data loss risk)

**Root cause**: `fs::remove_file(dest)` followed by `fs::rename(tmp, dest)` creates a window where neither file exists; a crash between the two calls loses the CSV permanently.

**Fix**: Write to a sibling temp file (`.filename.tmp-<pid>`), then call a single `fs::rename(tmp, dest)` — on POSIX this is atomic and replaces the destination without a gap.

---

### #47 — `select_gpus`/`select_gpu_ids` silently drops invalid GPU indices (UX)

**Root cause**: When `--gpu N` refers to a non-existent index the code fell through the match without inserting any GPU and without returning an error, causing subsequent operations to run on zero GPUs.

**Fix**: Both functions now return `Err` with a human-readable message when no GPU matches the given index/ID.

---

### #48 / #49 — No input validation on numeric CLI args (overflow / crash)

**Root cause** (`arg_help.rs`): All numeric arguments (`--vboost`, `--tlimit`, `--voltage_delta`, `--core`, `--memory`, `--plimit`, `--level`, `--timeout_loops`, etc.) were declared as raw strings and parsed at call sites with no range checks. Out-of-range values (e.g. `--core 999999`) caused integer overflows or panics deep in the NvAPI/NVML call path.

**Fix**: Added `value_parser!(T).range(lo..=hi)` to every numeric argument, letting clap reject bad values before any code runs. All call sites updated to use typed getters (`get_one::<u32>()`, `get_one::<i32>()`, etc.).

---

### #17 — Memory autoscan progress always shows 0.00% (UX)

**Root cause** (`oc_scanner.rs`): Integer division truncated the result to 0 before the `{:.2}%` format string could display decimals.

**Fix**: Promote to `f64` before dividing, multiply by 100.0, guard against divide-by-zero.

---

## Test plan

- [x] `cargo check` — zero errors, zero warnings
- [x] `cargo test` — 3/3 tests pass (`test_resolution_downgrade_ladder`, `test_resolution_r640x384_not_self_loop`, `test_resolution_r400x300_is_terminal`)
- [ ] Manual smoke-test: `nvoc set vfp autoscan --help` — verify range-bounded args visible in help
- [ ] Manual smoke-test: `nvoc set vfp autoscan --timeout_loops 0` — verify clap rejects out-of-range value
- [ ] Manual smoke-test: `nvoc --gpu 99 get` — verify error message for invalid GPU index

## Issues closed

Closes #17, #42, #43, #44, #46, #47, #48, #49

https://claude.ai/code/session_01UuB2RiU4fsy8TaYEGZP2r4

---
_Generated by [Claude Code](https://claude.ai/code/session_01UuB2RiU4fsy8TaYEGZP2r4)_

### DIFF
--- a/auto-optimizer/src/arg_help.rs
+++ b/auto-optimizer/src/arg_help.rs
@@ -6,7 +6,7 @@ use crate::platform::{
 use crate::types::{
     OutputFormat, ResetSettings, VfpResetDomain, POSSIBLE_BOOL, POSSIBLE_BOOL_OFF, POSSIBLE_BOOL_ON,
 };
-use clap::{Arg, ArgAction, Command};
+use clap::{value_parser, Arg, ArgAction, Command};
 use nvapi_hi::PState;
 
 pub fn get_arguments() -> Command {
@@ -199,18 +199,20 @@ pub fn get_arguments() -> Command {
                                   .short('c')
                                   .long("core")
                                   .value_name("CORE_MHZ")
-                                  .help("Target absolute core frequency in MHz")
+                                  .help("Target absolute core frequency in MHz (100–5000)")
                                   .num_args(1)
                                   .required(true)
+                                  .value_parser(value_parser!(u32).range(100..=5000))
                           )
                           .arg(
                               Arg::new("memory")
                                   .short('m')
                                   .long("memory")
                                   .value_name("MEM_MHZ")
-                                  .help("Target absolute memory frequency in MHz")
+                                  .help("Target absolute memory frequency in MHz (100–5000)")
                                   .num_args(1)
                                   .required(true)
+                                  .value_parser(value_parser!(u32).range(100..=5000))
                           )
                   )
                   .subcommand(
@@ -221,10 +223,11 @@ pub fn get_arguments() -> Command {
                             Arg::new("vboost")
                                 .short('V')
                                 .long("voltage-boost")
-                        .value_name("VBOOST")
-                        .num_args(1)
-                        .help("Voltage Boost %"),
-                )
+                                .value_name("VBOOST")
+                                .num_args(1)
+                                .value_parser(value_parser!(u32).range(0..=200))
+                                .help("Voltage Boost % (0–200)"),
+                        )
                 .arg(
                     Arg::new("tlimit")
                         .short('T')
@@ -232,7 +235,8 @@ pub fn get_arguments() -> Command {
                         .value_name("TEMPLIMIT")
                         .num_args(1..)
                         .action(ArgAction::Append)
-                        .help("Thermal limit (C)"),
+                        .value_parser(value_parser!(i32).range(40..=110))
+                        .help("Thermal limit °C (40–110)"),
                 )
                 .arg(
                     Arg::new("plimit")
@@ -241,7 +245,8 @@ pub fn get_arguments() -> Command {
                         .value_name("POWERLIMIT")
                         .num_args(1..)
                         .action(ArgAction::Append)
-                        .help("Power limit %"),
+                        .value_parser(value_parser!(u32).range(0..=200))
+                        .help("Power limit % (0–200)"),
                 )
                 .arg(
                     Arg::new("voltage_delta")
@@ -250,7 +255,8 @@ pub fn get_arguments() -> Command {
                         .value_name("UV")
                         .num_args(1)
                         .allow_hyphen_values(true)
-                        .help("Core voltage delta in μV via SetPstates20 (for Maxwell/900-series and older, e.g. +100000 = +100mV). Target pstate selectable via -z."),
+                        .value_parser(value_parser!(i32).range(-250_000..=250_000))
+                        .help("Core voltage delta in μV via SetPstates20 (−250000..+250000). Target pstate selectable via -z."),
                 )
                 .arg(
                     Arg::new("pstate")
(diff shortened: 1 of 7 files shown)


Dixiao-L:
Heads up — overlap with several open / closed PRs noticed during a duplicate-branch sweep:

**Substantial overlap with #52** (same author, same files):
- Identical edits in `auto-optimizer/src/basic_func.rs` (TestResolution self-loop, `select_gpus`/`select_gpu_ids` error-on-miss) and `auto-optimizer/src/oc_profile_function.rs` (atomic CSV write, `linear_interpolate`, `get_key_points_indices`, `parse_col` helper, `key_point_extractor` empty-guard, `export_vfp_from_log` assert→warn).
- This PR adds: #17 (autoscan progress %), #48/#49 (clap `value_parser!.range()` validation), unit tests for `TestResolution::downgrade`.
- This PR is missing: **#45 DEVMODEW alignment fix** from #52 (`#[repr(C, align(4))] AlignedBuf`). If #52 closes in favor of this PR, that fix needs to be ported here or filed separately.

**Overlap with #24** (autoscan progress %): same fix in `oc_scanner.rs::run_mem_oc_phase`, slightly different style. #24 can close if this lands.

**Partial overlap with #25** (u32 overflow #16): the `main.rs` range check is replaced by `value_parser!(u32).range(100..=5000)` in `arg_help.rs` here — good. However #25 also adds a `saturating_mul` backstop in `oc_get_set_function_nvapi.rs::set_legacy_clocks_nvapi` that this PR does **not** have. Worth keeping that backstop as defense-in-depth.

No action requested — just flagging so you can decide which PRs to consolidate.

---

Dixiao-L:
This PR's fixes (#17, #42–#44, #46–#49) are fully included in the consolidated #57, which also adds #15, #18, #45, #50, #51, #53. #57 has the same `cargo check` passing guarantee. Suggest closing this one once #57 is reviewed to keep a clean PR queue.

---

Dixiao-L:
Superseded by #57, which covers the same fixes (#17, #42–#49) along with additional root-cause fixes. Suggest closing in favour of #57 (for the auto-optimizer bugs) and #58 (for the comprehensive arg_help.rs range validators that #49 specifically requested).

---

Dixiao-L:
Superseded by #57 — strict subset (#17, #42–#49 are all included in the consolidated 14-bug PR; #49 is additionally addressed at the clap layer in #58). Closing as part of PR-queue triage.

---

PR #55, opened 2026-05-09
fix(auto-optimizer,srv): panics, silent GPU drops, unsafe FFI, atomic writes, NVML overflow, srv build (closes #42–#48, #50, #51, #53)

Dixiao-L:
## Summary

Fixes ten bugs across `auto-optimizer` and `srv`. All changes verified with `cargo check` (zero errors, only pre-existing dead-code warnings).

---

### Issue #45 — `get_primary_screen_size_raw` uses 1-byte-aligned `[u8; 220]` for `DEVMODEW`

**Root cause:** `DEVMODEW` DWORD fields require 4-byte alignment. The raw `[u8; 220]` buffer is 1-byte aligned, violating the unsafe contract and tripping sanitizers.

**Fix:** `#[repr(C, align(4))] struct AlignedBuf([u8; 256])` — compiler-enforced 4-byte alignment; 256 bytes provides headroom; `dmSize` still advertises 220 to the kernel.

---

### Issue #46 — `TestResolution::R640x384.downgrade()` self-loop hangs autoscan

**Root cause:** `R640x384 => Some(R640x384)` — unfinished edit. `R400x300 => Some(R400x300)` violates the `Option` contract.

**Fix:** `R640x384 => Some(R576x360)`, `R400x300 => None`. Two-line change.

---

### Issue #47 — `select_gpus` / `select_gpu_ids` silently drop invalid `--gpu` entries

**Root cause:** Both functions silently `continue`d past any unmatched GPU index.

**Fix:** Replace silent skip with `Error::Custom("no GPU matches --gpu N; use \`nvoc list\` to see available indices")` — any invalid index is now a hard error.

---

### Issue #44 — `update_csv_with_load_and_margin` non-atomic CSV replace loses user data

**Root cause:** Temp file written to hardcoded `./ws/temp.csv`, then `remove_file` + `rename` — crash in the gap deletes the user's CSV permanently.

**Fix:** Write temp file as `.filename.tmp-<pid>` next to destination (same filesystem). Drop explicit `remove_file` — `fs::rename` over an existing path is atomic on POSIX.

---

### Issue #43 — `linear_interpolate` divide-by-zero and `u32` underflow

**Root cause:** `(v2-v1)` could be zero yielding `NaN`, and `(current_v-v1)` unsigned subtraction underflowed when `current_v < v1`.

[...]

Issue #53, opened 2026-05-09
auto-optimizer: `autoscan_gpuboostv3` ultrafast key-point fix-up underflows `usize` on small key-point indices


Dixiao-L:
## Summary

`autoscan_gpuboostv3` adjusts the four ultrafast key-points (`p1..p4`) by adding/subtracting fixed offsets, treating them as `usize`. When a key-point lands at index 0..5 (entirely possible if the autoscan log is fresh, the V-F curve is short, or `key_point_extractor` returns 0 as a fallback), the subtractions underflow `usize`. In debug builds this panics outright. In release builds it wraps to `usize::MAX`-adjacent values which then propagate through downstream comparisons in surprising ways.

The release-build behavior happens to "work" for one site (the underflowed value is huge, makes the `<` comparison true, and the if-body sets `p1` to a sane value), but two other sites do not have that recovery — and "lucky correctness via wrap" is not a property the autoscan should depend on.

## Evidence

`auto-optimizer/src/oc_scanner.rs:1375-1396`:

```rust
if p1 - 6 < lower_voltage_point {                  // ← underflow if p1 < 6
    p1 = lower_voltage_point + 6;
}
if p2 < lower_voltage_point {
    p2 = lower_voltage_point + 10;
}
if p3 > upper_voltage_point {
    p3 = upper_voltage_point - 10;                 // ← underflow if upper < 10
}
if p4 > upper_voltage_point {
    p4 = upper_voltage_point;
}

// stair bias
if is_50_series && p1 == p2 {
    p1 -= 6;                                        // ← underflow if p1 < 6
    p2 += 6;
}
if is_50_series && p2 == p3 {
    p2 -= 6;                                        // ← underflow if p2 < 6
    p3 += 6;
}
```

Where `p1, p2, p3, p4: usize` and originate from `key_point_extractor` (`oc_profile_function.rs:920`) which can return `(0, 0, 0, 0)` via the `unwrap_or(0)` fallbacks at `oc_profile_function.rs:986-991`.

Concrete failure paths:

- **Line 1375:** `p1 - 6` with `p1 < 6` underflows. In debug: panic. In release: wraps to ~`usize::MAX`, which is greater than `lower_voltage_point` (typically 40-70), so the `if`-body fires and sets `p1` to a sane value. Currently "lucky" but undefined behavior to depend on.
[...]

Issue #51, opened 2026-05-09
srv: build broken — missing handle_unlock_vfp / handle_global_oc_offset_subcommand in nvoc_auto_optimizer


Dixiao-L:
## Summary

`srv/` cannot compile against `auto-optimizer/` at HEAD. Two functions imported by `srv/src/bin/nvoc_service.rs` no longer exist in the `nvoc_auto_optimizer` library:

- `handle_unlock_vfp`
- `handle_global_oc_offset_subcommand`

`handle_lock_vfp` and `find_matching_vfp_point` still exist in `auto-optimizer/src/oc_get_set_function_nvapi.rs`; the other two were removed (or renamed) without updating the `srv` consumer.

## Reproduction

```
cd srv && cargo build --release
```

Fails on hosted Windows CI; will fail on any developer machine targeting Windows.

## Impact

- `srv` is unbuildable. The Windows service / HTTP control endpoint cannot be produced from `main`.
- CI gating for `srv` (#19) had to be set to `continue-on-error: true` to unblock the CI bootstrap. The job will keep showing as failed (yellow) until this is fixed.

## Proposed fixes (pick one)

1. **Reinstate the missing symbols** in `auto-optimizer/` with their previous behaviour (preferred if they were removed accidentally — check `git log -p auto-optimizer/src/oc_get_set_function_nvapi.rs`).
2. **Update `srv`** to call whatever the replacements are (e.g. if `handle_global_oc_offset_subcommand` was split or renamed, port the call site).
3. If the functionality is intentionally gone, **rip the dead code paths out of `srv/src/bin/nvoc_service.rs`** so the binary builds without those imports.

## Follow-up

Once fixed, drop `continue-on-error: true` from the `rust-srv` job in `.github/workflows/ci.yml` and add a `cargo test --no-fail-fast -- --skip gpu_` step (currently no tests exist in `srv/`).

Related: PR #19 (CI bootstrap).

Dixiao-L:
Confirmed by direct grep against `auto-optimizer/src/` at HEAD:

```
$ grep -nE 'fn handle_unlock_vfp|fn handle_global_oc_offset_subcommand' auto-optimizer/src/*.rs
(no matches)

$ grep -nE 'fn handle_lock_vfp|fn find_matching_vfp_point' auto-optimizer/src/*.rs
auto-optimizer/src/oc_get_set_function_nvapi.rs:555:pub fn handle_lock_vfp(
[...]

Issue #50, opened 2026-05-09
auto-optimizer: `voltage_frequency_check` panics on `status.voltage = None` — unguarded unwrap in autoscan hot path


Dixiao-L:
## Summary

`voltage_frequency_check` (called from inside the autoscan inner loop and from the soft-temperature worker in `srv`) unwraps `status.voltage` directly. If NvAPI doesn't return a voltage in the status payload — which happens on some GPU/driver combinations and on transient driver wobble — the function panics, taking down the autoscan or the service.

The function already returns `Result<bool, Error>`, so the correct shape is `?` propagation. The unwrap is inconsistent with the rest of the function (every other field access uses `.ok_or(...)?`).

## Evidence

`auto-optimizer/src/oc_get_set_function_nvapi.rs:1256-1313`:

```rust
pub fn voltage_frequency_check(arg_matches: ArgMatches, point: usize) -> Result<bool, Error> {
    let gpu = arg_matches.get_many::<String>("gpu");
    let gpu_list = crate::basic_func::get_sorted_gpus()?;
    let gpus = select_gpus(&gpu_list, gpu)?;
    let mut precise_flag = false;

    for gpu in gpus {
        let status = gpu.status()?;
        let readout_v = status.voltage.unwrap();              // ← panic on None
        let readout_f = status.clone().clocks;
        ...

        let current_point = status.clone().vfp.ok_or(Error::VfpUnsupported)?.graphics;

        let default_v = current_point
            .get(&(point))
            .ok_or(Error::Str("invalid point index"))?
            .voltage;
        ...
```

Note the contrast on the very next lines: `status.vfp.ok_or(Error::VfpUnsupported)?` and `current_point.get(&(point)).ok_or(Error::Str("invalid point index"))?` are both proper `?` propagation. The `status.voltage.unwrap()` is the outlier.

`status.voltage` is an `Option<Microvolts>` because NvAPI sometimes omits it — e.g. on professional cards that don't expose voltage telemetry, on GPUs in low-power states where the voltage sensor is gated, or during a brief window after a driver TDR.

## Why it matters

`voltage_frequency_check` is called from at least three hot paths in `oc_scanner.rs` (lines 258, 510, 847, 989, 1102) — every one of those is inside the autoscan stress-test loop. A transient `None` voltage during a stress run panics the entire `auto-optimizer` process. The user loses the run; the next-step recovery never runs because the process is dead.

[...]

PR #57, opened 2026-05-09
fix(auto-optimizer,srv): 14 root-cause fixes — panics, srv build, overflows, data corruption, atomic writes (closes #15–#18, #42–#48, #50–#51, #53)

### LINKED ISSUE #15: auto-optimizer: `pre_load_vf_recheck` panics whole autoscan on a recoverable v-f read failure

Dixiao-L:
## Summary

Consolidated root-cause fixes for 14 open bugs across `auto-optimizer` and `srv`. This PR supersedes #24, #25, #28, #30, #52, #54, #55. All verified: `cargo check` passes on both crates (zero errors, only pre-existing dead-code warnings).

---

### #46 — `TestResolution::R640x384.downgrade()` self-loop hangs autoscan
**Root cause:** `R640x384 => Some(R640x384)` and `R400x300 => Some(R400x300)` — incomplete match arms; any resolution-downgrade loop spun forever.  
**Fix:** `R640x384 => Some(R576x360)`, `R400x300 => None`.

---

### #47 — `select_gpus` / `select_gpu_ids` silently drop invalid `--gpu` entries
**Root cause:** All three lookup paths fell through to a silent `continue` on no match; `--gpu 0,99` on a 1-GPU system returned only GPU 0 with no warning.  
**Fix:** Any unmatched input immediately returns `Err(Error::Custom("no GPU matches --gpu N; use \`nvoc list\`…"))`.

---

### #45 — `get_primary_screen_size_raw` uses 1-byte-aligned buffer for `DEVMODEW`
**Root cause:** `DEVMODEW` DWORD fields require 4-byte alignment; `[u8; 220]` stack buffer violates the unsafe FFI contract.  
**Fix:** `#[repr(C, align(4))] struct AlignedBuf([u8; 256])` — compiler-enforced, `dmSize` still 220.

---

### #51 — `srv` build broken: `handle_unlock_vfp` / `handle_global_oc_offset_subcommand` missing
**Root cause:** `nvoc_service.rs` imported both functions; neither existed in `nvoc_auto_optimizer`.  
**Fix:** Both added to `oc_get_set_function_nvapi.rs`. **srv now compiles.**

---

### #50 — `voltage_frequency_check` panics on `status.voltage = None`
**Root cause:** `status.voltage.unwrap()` in autoscan hot path panics on professional GPUs or transient driver wobble.  
**Fix:** `.ok_or_else(|| Error::Custom(…))?` — recoverable error, not process kill.

---

### #48 — NVML power-limit and memory-clock setters overflow on user input
**Root cause:** `limit_w * 1000` (`u32`) and `offset * 2` (`i32`) wrap in release builds.  
**Fix:** `saturating_mul(1000)` and `saturating_mul(2)`.

---

### #16 — `set_legacy_clocks_nvapi` overflows on user MHz input before unsafe FFI
**Root cause:** `info.clocks[8] = mem_mhz * 1000` and `info.clocks[30] = core_mhz * 2000` — plain `u32` arithmetic writes garbage MHz to the undocumented `NvAPI_GPU_SetClocks` struct on overflow.  
**Fix (two layers):** `main.rs` range-guards `--core` / `--memory` to 100–5000 MHz; `saturating_mul(1000)` / `saturating_mul(2000)` as backstop at the assignment site.

---

### #53 — `autoscan_gpuboostv3` ultrafast key-point fix-up underflows `usize`
**Root cause:** `p1 - 6`, `p2 - 6`, `upper_voltage_point - 10` — plain `usize` subtraction panics in debug when indices are small.  
**Fix:** All four sites use `saturating_sub`.

---

### #17 — Memory autoscan progress always shows `0.00%`
**Root cause:** `usize / usize` integer division truncated to 0 before `{:.2}%` format.  
**Fix:** `as f64 / …max(1) as f64 * 100.0`.

---

### #18 — Autoscan fluctuation timer uses `SystemTime` (wall-clock, can go backwards)
**Root cause:** NTP correction made `.elapsed()` fail; the `unwrap_or(6 s)` fallback was larger than the 1.5 s trigger — every loop iteration fired the fluctuation step.  
**Fix:** `Instant::now()` (monotonic, infallible). Removed `unwrap_or`.

---

### #15 — `pre_load_vf_recheck` panics whole autoscan on recoverable v-f read failure
**Root cause:** `.expect("Failed to read v-f info")` kills the entire multi-hour scan on one transient NvAPI error.  
**Fix:** `-> Result<(), Error>` + `?` at all three call sites.

---

### #43 — `linear_interpolate` divide-by-zero and `u32` underflow write wrong deltas
**Root cause:** (1) `(v2-v1)` = 0 on flat V-F region → NaN → delta silently zeroed. (2) `(current_v-v1)` underflows as `u32` → absurdly large interpolated delta.  
**Fix:** Returns `Result`. Flat region returns midpoint via `i64`. Endpoints sorted; `current_v` bounds-checked.

---

### #44 — `update_csv_with_load_and_margin` non-atomic CSV replace + hardcoded `./ws/` path
**Root cause:** Temp file at hardcoded `./ws/temp.csv`; `remove_file` + `rename` gap can permanently delete the user's CSV on crash.  
**Fix:** Sibling temp file `.filename.tmp-`; single atomic `fs::rename` replaces both steps.

---

### #42 — Multiple `panic!` / `.unwrap()` / `usize` underflow on bad CSV inputs
**Root cause:** Five panic sites in the `fix_result` / `export_log` / `key_point_extractor` workflow on user-supplied files.  
**Fix:** `parse_col<T>` helper; `get_key_points_indices` → `Result`; `interpolate_deltas` → `Result`; `key_point_extractor` guards `is_empty()`; `export_vfp_from_log` replaces `assert_eq!` with warning+`continue`; `extract_default_frequencies` uses safe `.get(N).ok_or_else(…)?`.

---

## Test plan

- [x] `cargo check` (auto-optimizer, dev) — zero errors
- [x] `cargo check` (srv, dev) — zero errors, unbreaks #51
- [ ] Live Windows: `nvoc --gpu 0,99 status` on 1-GPU system → hard error message
- [ ] Live Windows: autoscan at R640x384 → steps down R576x360 → R400x300 → stops
- [ ] Live Windows: `fix_result --ultrafast` against CSV with 3 key points → clean error
- [ ] Live Windows: `fix_result --ultrafast` against equal-voltage rows → non-zero deltas
- [ ] Live Windows: memory autoscan phase → progress increments smoothly 0%→100%
- [ ] Live Windows: autoscan with NTP clock slip mid-run → fluctuation rate stays ~0.67 Hz
- [ ] Live Windows: `nvoc set legacy-clock --core 50` → rejected with range error
- [ ] Live Windows: `nvoc set legacy-clock --core 800 --memory 1000` → succeeds

## Relationship to other PRs

Supersedes (suggest closing): #24 (#17), #25 (#16), #28 (#15), #30 (#18), #52 (#42–#47), #54 (#17,#42–#49), #55 (#42–#48,#50,#51,#53).  
Stressor bugs (#33–#38) are tracked in #40/#41 — out of scope here.

https://claude.ai/code/session_01CwT6EB9Lm18bxVzJwY4WgX


84634E1A607A (dismissed):
我看不懂！

auto-optimizer/src/oc_profile_function.rs
```diff
@@ -214,12 +218,19 @@ fn update_csv_with_load_and_margin(
     minimum_delta_core_freq_step: i32,
     legacy_flag: bool,
 ) -> Result<(), Error> {
+    let dest_path = std::path::Path::new(file_path);
+    let tmp_name = format!(
+        ".{}.tmp-",
+        dest_path.file_name().and_then(|n| n.to_str()).unwrap_or("tmp")
+    );
+    let tmp_path = dest_path.parent().unwrap_or(dest_path).join(&tmp_name);
+
```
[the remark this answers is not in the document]
Dixiao-L: Fixed. The temp filename now includes the process ID: `.{filename}.{pid}.tmp`, which guarantees uniqueness across concurrent runs and across crash survivors on the same machine. See commit `fec4b9b` (pushed to this branch).

---

auto-optimizer/src/oc_profile_function.rs
```diff
@@ -265,8 +276,7 @@ fn update_csv_with_load_and_margin(
     wtr.flush()?;
 
     // Replace original file with updated file
-    fs::remove_file(file_path)?;
-    fs::rename("./ws/temp.csv", file_path)?;
+    fs::rename(&tmp_path, file_path)?;
```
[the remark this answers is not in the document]
Dixiao-L: Fixed. The replacement now tries an atomic `fs::rename` first (which works on Linux and modern Windows); if that fails (older Windows where rename-over-existing is rejected), it explicitly removes the destination and retries, cleaning up the tmp file on any error so it is never silently lost. See commit `fec4b9b`.

---

Dixiao-L:
@codex

auto-optimizer/src/basic_func.rs
```diff
@@ -146,11 +149,9 @@ impl TestResolution {
             TestResolution::R800x600 => Some(TestResolution::R768x576),
             TestResolution::R768x576 => Some(TestResolution::R720x480),
             TestResolution::R720x480 => Some(TestResolution::R640x384),
-            TestResolution::R640x384 => Some(TestResolution::R640x384),
-            // TestResolution::R640x384 => Some(TestResolution::R576x360),
+            TestResolution::R640x384 => Some(TestResolution::R576x360),
             TestResolution::R576x360 => Some(TestResolution::R400x300),
```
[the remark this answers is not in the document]
Dixiao-L: Fixed. The hard-coded `resolution != R640x384` sentinel is replaced with `resolution.downgrade().is_some()`, which derives the terminal condition directly from the resolution chain. With this change:

* `R640x384` → `downgrade()` = `Some(R576x360)` → flag set → downgrade proceeds
* `R576x360` → `downgrade()` = `Some(R400x300)` → flag set → downgrade proceeds
* `R400x300` → `downgrade()` = `None` → flag not set → loop exits cleanly (true terminal)

The now-unused `R640x384` import is also removed. See commit `6f102ef`.

---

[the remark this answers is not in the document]
Dixiao-L: This was already fixed before the Codex review was posted. The review ran on commit `c237d4c635`; the branch HEAD is now `6f102ef`.

The current `oc_scanner.rs` (lines 327 and 332) uses `resolution.downgrade().is_some()` as the throttle gate — not a hardcoded `resolution != R640x384` comparison. The `downgrade()` chain in `basic_func.rs` was fixed in this same PR:

```rust
TestResolution::R640x384 => Some(TestResolution::R576x360),
TestResolution::R576x360 => Some(TestResolution::R400x300),
TestResolution::R400x300 => None,
```

All three new rungs (`R640x384 → R576x360 → R400x300`) are fully reachable in the automatic downgrade path.

---

auto-optimizer/src/oc_get_set_function_nvapi.rs
```diff
@@ -1583,3 +1582,51 @@ pub fn set_legacy_clocks_nvapi(gpu: &Gpu, core_mhz: u32, mem_mhz: u32) -> Result
...
+        .transpose()?
+        .unwrap_or(ClockDomain::Graphics);
+
+    for gpu in gpus {
+        gpu.inner()
+            .set_pstates([(pstate, domain, KilohertzDelta(delta))].iter().cloned())
+            .map_err(Error::from)?;
+    }
+    Ok(())
+}
```
physicsdolphin: see https://github.com/Skyworks-Neo/nvoc/pull/56#pullrequestreview-4259173578 and https://github.com/Skyworks-Neo/nvoc/pull/56#issuecomment-4415421004 
dupe
do not add back function
other modification to this file is acceptable
@claude @Dixiao-L

auto-optimizer/src/basic_func.rs
```diff
@@ -146,11 +149,9 @@ impl TestResolution {
...
-            TestResolution::R400x300 => Some(TestResolution::R400x300),
-            // Lowest resolution, no downgrade available
+            TestResolution::R400x300 => None,
```
physicsdolphin, line 154: If resolution goes to None then stress load cannot be init. You have to admit that in some edge case even lowest resolution of 3D load will still hit power limit, then you always need a bottom line that cannot go down anymore @claude @Dixiao-L

auto-optimizer/src/oc_get_set_function_nvapi.rs
Dixiao-L: Fixed in `cbbf0ff`. Both `handle_unlock_vfp` and `handle_global_oc_offset_subcommand` are removed from `oc_get_set_function_nvapi.rs`. As established in PR #56, the srv path already inlines the VFP-unlock and global-OC-offset logic directly in `nvoc_service.rs` (using `g.reset_vfp_lock()`, `reset_vfp_frequency_lock`, and `g.inner().set_pstates(...)` respectively), so there is no call site left for the two functions. `cargo check` on both `auto-optimizer` and `srv` is clean.

---

auto-optimizer/src/basic_func.rs
Dixiao-L: Agreed — and the current code already upholds that invariant. At `R400x300` (where `downgrade()` returns `None`), the throttle gate on line 327 (`resolution.downgrade().is_some()`) evaluates to `false`, so `thrm_or_pwr_limit_flag` is never set (or, if it carried over from the previous `R576x360` iteration, line 332's `let Some(lower_res) = resolution.downgrade()` still won't match), and the function falls through to `return (exit_code, resolution)` — accepting the `R400x300` result as the floor regardless of throttle ratio.

Commit `cbbf0ff` adds an explicit `eprintln!` warning at this code path so the "accepting floor result under throttling" decision is observable in logs:

```
Warning: Thermal/power throttling detected (38%) at lowest resolution (R400x300).
No lower resolution available; accepting result as floor.
```

`R400x300` can never be set to `None` during the scan loop — the only way `resolution` is updated is via `let Some(lower_res) = resolution.downgrade()`, so the variable always holds a valid `TestResolution` variant.

---

84634E1A607A: Are we still relying on GUI stress test anyway? I don't think so?

84634E1A607A: I'd use another PR to completely remove the GUI path

84634E1A607A (approved):

Dixiao-L:
## Review thread verification (all 5 threads addressed) ✅

Verified the current HEAD (`cbbf0ff`) against every open review thread. CI re-ran after PR #77 merged to main — **all checks green**.

---

### Thread 1 & 2 — Copilot: temp filename uniqueness + Windows `fs::rename`

`update_csv_with_load_and_margin` (`oc_profile_function.rs`):

- Temp name is now `.{filename}.{pid}.tmp` — unique across concurrent runs and crash survivors.
- Rename is attempted first; if it fails on Windows (rename-over-existing rejected), the destination is removed and rename is retried; on any error the tmp file is also cleaned up. **No data loss path.**

### Thread 3 — Codex: resolution downgrade chain unreachable

Downgrade chain in `basic_func.rs` is:
```
R640x384 → R576x360 → R400x300 → None
```
`oc_scanner.rs` uses `resolution.downgrade().is_some()` as the throttle gate, making all three rungs fully reachable. Thread marked **outdated** by GitHub (code changed before the Codex review ran).

### Thread 4 — @physicsdolphin: dupe functions in `oc_get_set_function_nvapi.rs`

`handle_unlock_vfp` and `handle_global_oc_offset_subcommand` are confirmed absent from the file on HEAD — grep returns no matches. Fix matches the guidance from PR #56.

### Thread 5 — @physicsdolphin: "if resolution goes to None, stress load cannot init"

**Confirmed correct floor behavior.** Read the full `pressure_runner::run` function:

- `thrm_or_pwr_limit_flag` and all loop variables are declared **inside** the outer `loop {}` block (lines 226–232), so they reset to zero/false on every new process spawn — there is no carry-over from a previous R576x360 iteration.
- At R400x300 (`downgrade()` → `None`): the `if resolution.downgrade().is_some()` branch is **not taken**, `thrm_or_pwr_limit_flag` stays `false`, a warning is printed, and the function returns `(exit_code, R400x300)` — R400x300 is always the hard floor.

@physicsdolphin please resolve thread 5 (and thread 4 is already outdated) when you've confirmed this is satisfactory.


---

PR #80, opened 2026-05-10
Comment out checkout ref configuration

physicsdolphin:
Comment out the 'with' section in the checkout step


G4614 (approved):
fix approved

Issue #74, opened 2026-05-10
auto-optimizer: `export_single_point` leaks a `String` per matching CSV row via `Box::leak`


Dixiao-L:
## Where

`auto-optimizer/src/oc_profile_function.rs:138`, inside the `for line in reader.lines()` loop in `export_single_point`.

## What

```rust
for line in reader.lines() {
    let line = line?;
    let mut columns: Vec<&str> = line.split(',').collect();
    ...
    if columns.first() == Some(&&*voltage) && columns.len() > 3 {
        columns[2] = &delta;
        let y_value: i32 = columns[2].parse().unwrap_or(0);
        let col3_value: i32 = columns[3].parse().unwrap_or(0);
        let sum = y_value + col3_value;
        columns[1] = Box::leak(sum.to_string().into_boxed_str());   // <-- leaks per matching row
    }
    record_lines.push(columns.join(","));
}
```

`Box::leak` is used to coerce a freshly-allocated `String` into a `&'static str` so it can be assigned to `columns[1]` (which has the borrow lifetime of `line`). The `columns` `Vec` is dropped at the end of the loop iteration once `record_lines.push(columns.join(","))` runs — but the `Box::leak`'d allocation lives for the lifetime of the process.

## Why it matters

- Every matching voltage row leaks ~10–20 bytes; on a 100-row VFP table with multiple matches per scan, each `set vfp single_point_adj` call leaks dozens of strings.
- The CLI is intended to be invoked repeatedly by the GUI/TUI during a long autoscan run, so the leak compounds across the lifetime of the parent process if the auto-optimizer is ever embedded as a library (per the architectural direction in #3).
- The pattern is also a code smell: `columns[1] = &sum.to_string()` would suffice if `columns` were `Vec<String>`, removing the lifetime hack entirely.

## Suggested direction

Change `columns: Vec<&str>` to `Vec<String>` (or `Vec<Cow<'_, str>>`) inside this function so the new value can be assigned without any unsafe-lifetime gymnastics. No `Box::leak` needed. While there, the same loop also has `unwrap_or(0)` on parse — fine for now, but worth a comment that silently zeroing malformed rows is intentional rather than a TODO.

## Severity

Low — slow leak, single-shot CLI process means current users barely notice. Medium if/when the auto-optimizer is consumed as a library (#3, #4).

## Not a duplicate of

[...]

Issue #73, opened 2026-05-10
srv: NVML/NVAPI device-count drift in nvoc_service panics on `gpu_dynamic_lock_point[i]`


Dixiao-L:
## Where

`srv/src/bin/nvoc_service.rs` — `run_service`, lines ~184–322.

## What

The dynamic VFP lock-point vector is sized from the **NVAPI** enumeration but indexed with the **NVML** iteration variable:

```rust
let nvml = Nvml::init().unwrap();
let gpus = Gpu::enumerate().unwrap();                                    // NVAPI
let vfp_highest_lock_point = 100;
let mut gpu_dynamic_lock_point: Vec<usize> =
    vec![vfp_highest_lock_point; gpus.len()];                            // sized by NVAPI

...

let count = nvml.device_count().unwrap_or(0);                            // NVML count
for i in 0..count {
    ...
    // OOB panic when `count > gpus.len()`
    if gpu_dynamic_lock_point[i as usize] >= vfp_highest_lock_point { ... }
    let current = gpu_dynamic_lock_point[i as usize];
    let next = current.saturating_sub(1).max(vfp_lowest_lock_point);
    gpu_dynamic_lock_point[i as usize] = next;
    ...
}
```

`count` and `gpus.len()` are not guaranteed equal:
- NVAPI does not enumerate datacenter SKUs (Tesla, some Quadro/MIG configurations) that NVML does see.
- Driver/feature-flag differences (e.g. NVAPI restricted client) can hide GPUs from NVAPI but not NVML.
- Even when the counts match, ordering between the two backends is not guaranteed identical, so the lock state being mutated for "GPU i" may correspond to a different physical GPU than the one whose temperature was just read.

## Why it matters

- When `count > gpus.len()`: the service panics with index-out-of-bounds the first time it touches the missing slot. Combined with #67 (no panic hook), the control plane silently dies.
- When the order differs but counts match: the temperature read for one GPU drives the VFP lock for a different GPU — the opposite of what the dynamic temp control is supposed to do.

## Suggested direction

- Iterate by NVAPI handle (`gpus.iter().enumerate()`) and use NVML lookup-by-PCI-bus-id to fetch the per-GPU temperature for each NVAPI GPU, instead of iterating `0..nvml.device_count()`.
- Or: build a single `Vec<(NvapiGpu, NvmlDevice)>` joined on PCI bus id at startup, fail fast if the join is incomplete, and key the lock-point vector by that joined index.
[...]

Issue #68, opened 2026-05-10
auto-optimizer: `export_single_point` trusts `--output` / `--initcsv` paths verbatim — arbitrary file overwrite when run as admin/root


Dixiao-L:
## Where

`auto-optimizer/src/oc_profile_function.rs` `export_single_point` (~lines 67–79) and the surrounding `fs::copy(init_path, file_path)?` call.

## What

`--output` and `--initcsv` are pulled out of `ArgMatches` with `.unwrap()` and passed straight to `fs::copy` / `File::create` with no canonicalization, no workspace bound, and no rejection of absolute / `..` / UNC paths.

The CLI documents itself as requiring admin/root for writes, so any path the user (or a frontend caller, or a config file) supplies is used verbatim with elevated privileges.

## Why it matters

Same bug shape as #44 (which is about `update_csv_with_load_and_margin`'s hardcoded `./ws/temp.csv` rename), but in a different function. A misconfigured GUI flow, a malicious autoscan config, or a typo can overwrite arbitrary files on the host. The `.unwrap()` on missing args is a separate panic-on-input issue.

## Suggested direction

- Canonicalize both paths and assert they live under a designated workspace root.
- Reject absolute paths and any component containing `..`.
- Replace `.unwrap()` on the argument lookups with proper error returns (overlap with the broader unwrap cleanup in #42 / #50).

## Severity

High — file-overwrite primitive under admin.

Issue #67, opened 2026-05-10
srv: HTTP server thread has no panic hook — silent control-plane loss without SCM failure report


Dixiao-L:
## Where

`srv/src/bin/nvoc_service.rs` ~lines 158–160 (HTTP server thread spawn) and `srv/src/websrv.rs` (the `for request in server.incoming_requests()` loop, with several `.unwrap()` calls — see also #7).

## What

The HTTP control-plane thread is spawned with `thread::spawn(...)`. If the loop panics (e.g. `serde_json::to_string(&*cfg).unwrap()` at `websrv.rs:31` or the `request.respond(...).unwrap()` calls), the thread dies silently. The Windows service keeps running and reports `Running` to SCM, but the control endpoint is gone.

## Why it matters

A monitored service that has lost its only control surface but still reports healthy is the worst observability failure mode. Operators won't know overclock commands are being dropped on the floor.

## Suggested direction

- Wrap the HTTP loop in a thread that, on panic, transitions the service to `Stopped` / failure state via SCM (or restarts the listener with backoff).
- Replace the `.unwrap()` calls in `websrv.rs` with logged `?` propagation that returns a 500 to the caller (overlap with #7).

## Severity

Medium.

Dixiao-L:
Already fixed in `srv/src/bin/nvoc_service.rs` (lines 160–176). The HTTP server thread is now wrapped in a `catch_unwind` + restart loop: a panic logs an error and restarts after 1 s, so a silent control-plane loss is no longer possible. Closing.

---
_Generated by [Claude Code](https://claude.ai/code)_

Dixiao-L:
Confirmed fixed. `srv/src/bin/nvoc_service.rs` now wraps the HTTP server thread in a `std::panic::catch_unwind` + restart loop (see the comment "Restart the HTTP thread on any exit (panic or early normal return)"). A panic no longer silently kills the control plane — the thread respawns with a 1 s backoff. Closing.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #86, opened 2026-05-11
fix(srv,auto-optimizer): graceful error paths + typed DEVMODEW (closes #7, #45)

### LINKED ISSUE #7: srv: replace `.unwrap()` calls with proper error handling in service startup and HTTP loop
## Summary

```rust

    ...
```

### B. HTTP loop unwraps

```rust
```

### C. Count

```
srv/src/websrv.rs:6
```

## Why it matters

## Reproduction

## Recommended approach

## Out of scope

### LINKED ISSUE #66: srv: log directory derived from `current_exe().parent().parent()` with `canbedel-` prefix — leftover debug + DLL/exe-planting risk
## Where

`srv/src/bin/nvoc_service.rs` ~lines 73–85.

## What

The service log directory is derived at runtime as `env::current_exe().parent().parent()` (with a sibling-dir fallback). The log filename uses a hardcoded `canbedel-{service_name}-output.log` prefix, which reads as leftover debug naming.

## Why it matters

The service runs as SYSTEM. Anchoring its log path off `current_exe()` parents means whoever controls that directory tree controls where SYSTEM-owned logs land — combined with the existing pattern of resolving sibling binaries from the same directory (see `install_service.rs`), a writable install dir becomes an unsigned-binary / log-redirection foothold.

The `canbedel-` prefix is also user-visible noise that suggests a temporary file but isn't one.

## Suggested direction

- Hardcode the log root to a SYSTEM-only path (e.g. `%ProgramData%\nvoc\logs`) and verify ACLs at startup.
- Drop the `canbedel-` prefix; pick a stable name.
- (Discussion welcome here rather than in the PR.)

## Severity

Medium — depends on install posture; High if the service is ever installed under a user-writable path.

Dixiao-L:
## Summary

Two root-cause fixes for issues not covered by any existing open PR.

---

### #7 — srv: `.unwrap()` calls in service startup and event handler

**Root cause:** Multiple `.unwrap()` calls throughout `nvoc_service.rs` contradict the documented intent ("service can be stopped even if Nvml fails to initialize"). The most dangerous are:

- `shutdown_tx.send(()).unwrap()` inside the SCM event handler — panicking here leaves the Service Control Manager hanging mid-stop-request, with no `ServiceState::Stopped` ever reported.
- `service_control_handler::register(...).unwrap()` — a panic here exits `my_service_main` before the final `set_service_status(Stopped)`, leaving SCM in an inconsistent state.
- `Nvml::init().unwrap()` / `Gpu::enumerate().unwrap()` — panics here kill the async worker without the service ever becoming stoppable via SCM.
- `RuntimeBuilder::new().build().unwrap()` — same as above.
- `config.lock().unwrap()` in the timer loop — a poisoned mutex kills the periodic monitoring loop permanently.

**Fix (`srv/src/bin/nvoc_service.rs`):**

| Location | Before | After |
|---|---|---|
| Stop/UserEvent handlers | `shutdown_tx.send(()).unwrap()` | `let _ = shutdown_tx.send(())` |
| register() | `.unwrap()` | `match` → log + `return` |
| RuntimeBuilder | `.build().unwrap()` | `match` → log error |
| Nvml::init() | `.unwrap()` | `match` → log + `return Ok(())` |
| Gpu::enumerate() | `.unwrap()` | `match` → log + `return Ok(())` |
| config.lock() | `.unwrap()` | `.unwrap_or_else(\|e\| e.into_inner())` |

`nvml.device_by_index(i).unwrap()` inside the GPU timer loop is **intentionally deferred** — PR #72 already rewrites that loop for the NVML/NVAPI count-drift OOB guard (#73). Fixing both in parallel would cause a conflict; address it in the PR #72 rebase.

---

### #45 — auto-optimizer: DEVMODEW fixed-size buffer — alignment + hand-rolled offsets

**Root cause:** `get_primary_screen_size_raw` used a hand-rolled byte-offset map (`OFFSET_DM_SIZE = 68`, `OFFSET_PELS_WIDTH = 172`, …) and a raw `[u8; 220]` buffer. The original 1-byte alignment was already partially fixed (an `AlignedBuf` wrapper was added), but the offset map was still raw and unverified — the compiler had no knowledge of the struct layout.

**Fix (`auto-optimizer/src/basic_func.rs`):**

Replace the offset map and raw buffer with a typed `#[repr(C)] struct DevModeW` whose layout the compiler owns:

```rust
#[repr(C)]
struct DevModeW {
    dm_device_name: [u16; 32],  // 64 bytes
    dm_spec_version: u16,
    dm_driver_version: u16,
    dm_size: u16,               // offset 68 ← compiler-guaranteed
    dm_driver_extra: u16,
    dm_fields: u32,
    _display_union: [u8; 96],   // bytes 76–171
    dm_pels_width: u32,         // offset 172 ← compiler-guaranteed
    dm_pels_height: u32,
    _rest: [u8; 40],
}

const _: () = assert!(mem::size_of::<DevModeW>() == 220);
```

The compile-time assertion (`const _: () = assert!(...)`) locks in the 220-byte layout and will catch any future drift. No new crate dependencies — the `windows` crate is still deliberately avoided.

---

## Test plan

- [x] `git diff --stat` — only `srv/src/bin/nvoc_service.rs` and `auto-optimizer/src/basic_func.rs` touched
- [ ] `cargo check -p nvoc-srv` (Windows) — zero new errors or warnings
- [ ] `cargo check -p nvoc-auto-optimizer` (Windows) — `const _: ()` assertion passes at compile time
- [ ] Windows: rename/remove `nvml.dll`, start service → service logs "NVML init failed" and SCM shows `Stopped` cleanly (not the panic/hang of before)
- [ ] Windows: inject a failing `service_control_handler::register` → service logs error and exits; SCM does not hang
- [ ] Windows: send Stop to a running service → `ServiceState::Stopped` is always reported (no orphan-running regression)

## Relationship to other open PRs

- **PR #72** (`fix(auto-optimizer,srv): GPU selection + path-traversal + srv log/panic/OOB fixes`): Fixes #66, #67, #68, #73, #74. Modifies the GPU timer loop that still contains `nvml.device_by_index(i).unwrap()` (issue #7 residual). When #72 lands, the remaining unwrap should be addressed in its rebase.
- **PR #76** (`fix(gui,auto-optimizer): privilege preflight + POSIX single-instance + atomic config write`): Unrelated subsystems; no conflict.
- **PR #78** (`fix(cli-stressor-cuda): CPU fallback for legacy GPU cuBLAS kernel gaps`): Unrelated.

https://claude.ai/code/session_017Yy3CbpmjitGj8E5bT8W28

---
_Generated by [Claude Code](https://claude.ai/code/session_017Yy3CbpmjitGj8E5bT8W28)_


Dixiao-L:
**Superseded by PR #88** for the `nvoc_service.rs` (#7) portion.

**Note on the #45 overlap with PR #87:**

This PR and PR #87 both touch `basic_func.rs` for #45 but take different approaches:

| PR | Approach for #45 |
|----|-----------------|
| This PR (#86) | `#[repr(C)] struct DevModeW` (typed, no size increase) |
| PR #87 | Same typed struct approach + compile-time `assert!` on offsets |
| PR #88 | Heap allocation of `DEVMODEW_SIZE + u16::MAX` bytes (covers full `dmDriverExtra` range) |

PR #87's typed-struct approach has the advantage of compile-time layout verification. PR #88's heap-allocation approach fully eliminates the buffer-overflow risk for any driver's `dmDriverExtra` value.

**Recommendation:** merge PR #88 (complete #7 + real buffer-overflow fix) and PR #87 (complete #82 + typed-struct layout contract) — then close this PR (#86) and PR #85 and PR #84 which are all subsets. If the approach for #45 is contested, that discussion belongs in #88 vs #87 comments directly.

---

Dixiao-L:

PR #98 is the consolidated safety-fixes PR covering graceful error paths, typed DEVMODEW, and all other fixes from #84, #85, #87, #88, and #94 in a single clean commit. CI passed.

---

physicsdolphin (dismissed):

physicsdolphin (dismissed):

PR #85, opened 2026-05-11
fix(srv): proper error handling for 6 critical .unwrap() calls (closes #7)

Dixiao-L:
## Summary

Root-cause fix for #7. All six `.unwrap()` patterns that could silently kill the Windows service on production-realistic inputs are replaced with logged error handling.

**`websrv.rs` HTTP loop unwraps were already fixed in an earlier commit** — this PR only touches `nvoc_service.rs`.

---

### 1. SCM control handler — `shutdown_tx.send(()).unwrap()` (×2)

```rust
// Before
shutdown_tx.send(()).unwrap();

// After
let _ = shutdown_tx.send(());
```

If the async runtime has already exited and dropped `shutdown_rx`, the send returns `Err`. The old `.unwrap()` panicked inside the SCM control callback, leaving the Service Control Manager hanging on the Stop request. The fix drops the error silently — idempotent, as noted in the issue.

---

### 2. `service_control_handler::register` — `.unwrap()`

```rust
// After
let status_handle = match service_control_handler::register(SERVICE_NAME, event_handler) {
    Ok(h) => h,
    Err(e) => {
        error!("Failed to register service control handler: {:?}", e);
        return;
    }
};
```

Registration failure is now logged and exits `my_service_main` cleanly, allowing the `set_service_status(Stopped)` call at the bottom to run.

---

### 3. `RuntimeBuilder::new().build()` — `.unwrap()`

```rust
// After
match compio::runtime::RuntimeBuilder::new().build() {
    Ok(rt) => { let _ = rt.block_on(run_service(config, shutdown_rx, cmd_rx)); }
    Err(e) => error!("Failed to build async runtime: {}", e),
};
```

Runtime-build failure is logged; `set_service_status(Stopped)` still runs afterward.

---

### 4. `Nvml::init()` and `Gpu::enumerate()` — `.unwrap()` (×2)

```rust
// After
let nvml = match Nvml::init() {
    Ok(n) => n,
    Err(e) => {
        error!("NVML initialization failed: {:?}; service worker exiting cleanly", e);
        return Ok(());
    }
};
let gpus = match Gpu::enumerate() {
    Ok(g) => g,
    Err(e) => {
        error!("GPU enumeration failed: {:?}; service worker exiting cleanly", e);
        return Ok(());
    }
};
```

[...]

PR #84, opened 2026-05-11
fix(auto-optimizer): saturating_mul + voltage range guard in parse_lock_voltage (closes #82)

Dixiao-L:
## Summary

Follow-up to #16 (closed) and its "follow-up sweep" note. `parse_lock_voltage` was the **only** remaining multiplication in `oc_get_set_function_nvapi.rs` that still used bare `*` instead of `.saturating_mul()`.

- Fixes #82

---

### Root cause

`parse_lock_voltage` (line 381) contained:
```rust
input_voltage * 1000   // bare multiply
```
while every other mV→µV / MHz→kHz conversion in the same file uses `.saturating_mul()` — a pattern established by PR #57 (issue #16). The `< 10 000` guard prevents a runtime overflow today, but the inconsistency is a latent hazard for future refactors and an audit gap in the voltage-lock path.

Additionally, no upper-bound sanity check existed before the computed µV value was handed to `gpu.set_vfp_lock_voltage()` via the undocumented NVAPI path — so a user entering `--voltage 9999` (mV) would produce 9 999 000 µV ≈ **10 V**, which would be forwarded to the driver without rejection.

---

### Changes — `auto-optimizer/src/oc_get_set_function_nvapi.rs`

1. **`input_voltage * 1000` → `input_voltage.saturating_mul(1000)`** — style and overflow-hardening, consistent with all other conversions in the file.

2. **Range guard (0.5 V – 2.0 V)** — reject computed µV values outside a generous but sane GPU voltage window before reaching the NVAPI driver call:
   ```rust
   const MIN_LOCK_UV: u32 = 500_000;   // 0.5 V
   const MAX_LOCK_UV: u32 = 2_000_000; // 2.0 V
   if !(MIN_LOCK_UV..=MAX_LOCK_UV).contains(&voltage_uv) {
       return Err(Error::from(format!(
           "--voltage {} µV is outside the supported range {}–{} µV (0.5–2.0 V)",
           voltage_uv, MIN_LOCK_UV, MAX_LOCK_UV
       )));
   }
   ```

---

## Test plan

- [x] `cargo check -p nvoc-auto-optimizer` — clean
- [ ] `nvoc set nvapi --voltage 1000` (mV path) — accepted, converted to 1 000 000 µV
- [ ] `nvoc set nvapi --voltage 1000000` (µV path, ≥ 10 000) — accepted as-is
- [ ] `nvoc set nvapi --voltage 9999` — rejected: "9 999 000 µV is outside the supported range"
[...]

PR #87, opened 2026-05-11
fix(auto-optimizer): saturating_mul + range guard for parse_lock_voltage; typed DevModeW struct (closes #82, #45)

### LINKED ISSUE #16: auto-optimizer: `set_legacy_clocks_nvapi` overflows on user MHz input — no clamp before unsafe FFI

### LINKED ISSUE #82: auto-optimizer: `parse_lock_voltage` uses unchecked `* 1000` — style inconsistency + follow-up sweep from #16
## Summary

Follow-up to #16 (now closed). The issue body flagged a "follow-up sweep" of other write paths; this is the one remaining case found.

`parse_lock_voltage` in `auto-optimizer/src/oc_get_set_function_nvapi.rs` (line 381) is the **only** multiplication in the entire file that still uses bare `*` instead of `.saturating_mul()`:

```rust
// oc_get_set_function_nvapi.rs  line 378-382
let voltage_uv = if input_voltage >= 10_000 {
    input_voltage
} else {
    input_voltage * 1000   // ← bare multiply; every other conversion in this file uses .saturating_mul()
};
```

### Overflow risk (theoretical)
The `< 10_000` guard limits `input_voltage` to ≤ 9 999, so the product `9 999 × 1 000 = 9 999 000` fits in `u32` — **no runtime overflow today**. However:

- A future refactor that removes or widens the guard would silently introduce overflow.
- Clippy `-D clippy::integer_arithmetic` or similar lints flag bare `*` on `u32` even when provably in-range.
- The PR #57 that fixed #16 changed every _other_ identical pattern to `.saturating_mul()`; this one was missed.

### Semantic hazard
The `"point"` arg fed into `parse_lock_voltage` has **no clap range validator** when used in the `--voltage` path, so a user can enter any value < 10 000. At 9 999 mV the converted value is 9 999 000 µV ≈ **10 V** — far above any GPU's safe operating range. The function doesn't validate the result against a plausible voltage window before handing it to `gpu.set_vfp_lock_voltage()`.

## Evidence

```
auto-optimizer/src/oc_get_set_function_nvapi.rs

line 289-290  parse_lock_frequency      → .saturating_mul(1000)  ✓
line 399      parse_nvapi_locked_voltage_target → .saturating_mul(1000)  ✓
line 636-637  handle_lock_vfp (core)    → .saturating_mul(1000)  ✓
line 660-661  handle_lock_vfp (mem)     → .saturating_mul(1000)  ✓
line 569-570  set_nvapi_pstate_lock     → .saturating_mul(1000)  ✓
line 1546-47  set_legacy_clocks_nvapi   → .saturating_mul(1000)  ✓  (fixed by #16 / PR #57)

line 381      parse_lock_voltage        → * 1000                 ✗  ← only remaining case
```

## Recommended fix

**Minimal (style + overflow-hardening):**
```rust
// line 381
input_voltage.saturating_mul(1000)
```

**Full fix (also adds semantic guard):**
```rust
let voltage_uv = if input_voltage >= 10_000 {
    input_voltage
} else {
    input_voltage.saturating_mul(1000)
};
// Reject clearly out-of-range voltages before touching the driver
const MAX_LOCK_UV: u32 = 2_000_000; // 2 V — generous upper bound for any consumer GPU
const MIN_LOCK_UV: u32 = 500_000;   // 0.5 V
if !(MIN_LOCK_UV..=MAX_LOCK_UV).contains(&voltage_uv) {
    return Err(Error::from(format!(
        "--voltage {} µV is outside the supported range {}–{} µV",
        voltage_uv, MIN_LOCK_UV, MAX_LOCK_UV
    )));
}
```

## Severity

Low-to-medium. No user-visible bug in normal operation; the concern is code hygiene, future-refactor safety, and the lack of an upper-bound guard before an NVAPI voltage-lock call.

## Out of scope

- Range validation on `--locked-voltage`, `--locked-core-clocks`, `--locked-mem-clocks` (free-form string args that already use `.saturating_mul()` in their callers but have no clap range bounds) — same audit category, could be a separate follow-up.

Dixiao-L:
## Summary

Two safety/correctness fixes in `auto-optimizer`, both discovered during the post-#57 follow-up sweep.

---

### #82 — `parse_lock_voltage`: bare `* 1000` → `.saturating_mul()` + voltage range guard

**Root cause:** `parse_lock_voltage` in `oc_get_set_function_nvapi.rs` (line 381) was the **only** remaining `*` multiplication in the file — every other mV→µV conversion had already been moved to `.saturating_mul(1000)` by PR #57 (which fixed #16). This one was missed.

The `< 10_000` guard makes overflow unreachable today (max product = 9 999 × 1 000 = 9 999 000 < u32::MAX), but:
- A future refactor that widens or removes the guard would silently introduce overflow.
- The inconsistency trips Clippy's integer-arithmetic lints.
- There was **no upper-bound validation** before handing the value to `gpu.set_vfp_lock_voltage()` — a user could enter `9999` mV (≈ 10 V), far outside any GPU's safe operating range.

**Fix (`oc_get_set_function_nvapi.rs`):**
```rust
// before
input_voltage * 1000

// after
input_voltage.saturating_mul(1000)
// + range guard: 500 000 – 2 000 000 µV (0.5 V – 2.0 V)
const MIN_LOCK_UV: u32 = 500_000;
const MAX_LOCK_UV: u32 = 2_000_000;
if !(MIN_LOCK_UV..=MAX_LOCK_UV).contains(&voltage_uv) {
    return Err(Error::Custom(format!("--voltage {} µV is outside …", voltage_uv)));
}
```

---

### #45 — `get_primary_screen_size_raw`: typed `repr(C)` struct replaces byte-offset map

**Root cause:** `get_primary_screen_size_raw` in `basic_func.rs` held a hand-rolled byte-offset map for `DEVMODEW` fields (`OFFSET_DM_SIZE = 68`, `OFFSET_PELS_WIDTH = 172`, `OFFSET_PELS_HEIGHT = 176`) and accessed them via raw byte slices. No compiler-enforced layout contract existed. If offsets drifted from the true `DEVMODEW` layout (e.g., after a refactor or a layout misread), the function would silently read wrong memory.

A partial fix already existed (buffer bumped from 220 → 256 bytes, `align(4)` added), but the fragile byte-offset access remained.

**Fix (`basic_func.rs`):**

Defined `#[repr(C, align(4))] struct DevModeW` that mirrors `DEVMODEW` (wingdi.h, x86-64 ABI) with compiler-computed size and alignment:

```rust
#[repr(C, align(4))]
struct DevModeW {
    dm_device_name: [u16; 32],       // offset   0
    dm_spec_version: u16,             // offset  64
    dm_driver_version: u16,           // offset  66
    dm_size: u16,                     // offset  68
    dm_driver_extra: u16,             // offset  70
    dm_fields: u32,                   // offset  72
    _union_display_printer: [u8; 16], // offset  76
    dm_color: i16,                    // offset  92
    dm_duplex: i16,                   // offset  94
    dm_y_resolution: i16,            // offset  96
    dm_tt_option: i16,               // offset  98
    dm_collate: i16,                 // offset 100
    dm_form_name: [u16; 32],         // offset 102
    dm_log_pixels: u16,              // offset 166
    dm_bits_per_pel: u32,            // offset 168
    dm_pels_width: u32,              // offset 172
    dm_pels_height: u32,             // offset 176
    _rest: [u8; 40],                 // offset 180
}

// compile-time layout contract — fail at build time if layout drifts
const _: () = assert!(mem::size_of::<DevModeW>() == 220);
const _: () = assert!(mem::offset_of!(DevModeW, dm_size) == 68);
const _: () = assert!(mem::offset_of!(DevModeW, dm_pels_width) == 172);
const _: () = assert!(mem::offset_of!(DevModeW, dm_pels_height) == 176);
```

`dm_size` is set to `mem::size_of::<DevModeW>() as u16` instead of the hard-coded literal 220. FFI signature changed to `*mut DevModeW`. No new crate dependency.

---

## Test plan

- [x] `cargo check` (auto-optimizer) — zero new warnings or errors
- [ ] Windows: `nvoc set nvapi --voltage 9999` — must now error with range message instead of silently sending ~10 V to the driver
- [ ] Windows: `nvoc set nvapi --voltage 1000` (1000 mV = 1 000 000 µV, in range) — accepted
- [ ] Windows: `nvoc set nvapi --voltage 400` (400 µV, below MIN_LOCK_UV) — rejected
- [ ] Windows: autoscan resolution picker — correct screen size still reported (`get_primary_screen_size_raw` returns same values as before)

https://claude.ai/code/session_018UjLSAqbiTWpYw4HSfPaQR

---
_Generated by [Claude Code](https://claude.ai/code/session_018UjLSAqbiTWpYw4HSfPaQR)_


Dixiao-L:

PR #98 is the consolidated safety-fixes PR covering `saturating_mul` + voltage range guard (this PR's changes) along with all the other fixes from #84, #85, #86, #88, and #94. CI passed.

---

physicsdolphin (dismissed):

physicsdolphin (dismissed):

PR #88, opened 2026-05-11
fix(auto-optimizer,srv): DEVMODEW heap buffer + srv startup/loop .unwrap() removal (closes #45, #7)

### LINKED ISSUE #45: auto-optimizer: `get_primary_screen_size_raw` uses fixed 220-byte buffer for `DEVMODEW` — stack-corruption risk on extended driver layouts

Dixiao-L:
## Summary

Two root-cause safety fixes that are not covered by any existing open PR.

---

### #45 — `get_primary_screen_size_raw`: stack-corruption risk on extended driver layouts

**Root cause:** `EnumDisplaySettingsW` is called with a fixed 256-byte stack buffer (`AlignedBuf([u8; 256])`). The DEVMODEW base structure is 220 bytes, leaving only 36 bytes of headroom for `dmDriverExtra`. The `dmDriverExtra` field is a `u16`, meaning a driver may legally write up to 65 535 bytes of private data after the base — any driver with `dmDriverExtra > 36` would silently write past the end of the stack buffer.

**Fix (`auto-optimizer/src/basic_func.rs`):**
Remove `AlignedBuf`; allocate `vec![0u8; DEVMODEW_SIZE + u16::MAX as usize]` on the heap. This covers the full Win32-spec range (220 + 65 535 = 65 755 bytes) without any stack pressure, and removes the fixed-size alignment struct entirely. The pixel-dimension reads (`u32::from_le_bytes`) are alignment-safe because `Vec<u8>` provides byte-level access.

---

### #7 — srv: `.unwrap()` in service startup and monitoring loop

**Root cause:** `nvoc_service.rs` had 12 `.unwrap()` calls across `my_service_main` and `run_service`. Panics from most of these produce a silent process crash from SCM's perspective — the service transitions to `Stopped` with no log entry explaining why.

**Fix (`srv/src/bin/nvoc_service.rs`):**

#### `my_service_main` (returns `()` — cannot propagate errors)

| Site | Old | New | Rationale |
|------|-----|-----|-----------|
| `create_dir_all` | `.unwrap()` | `if let Err(e)` + `eprintln!` | Log dir failure shouldn't abort service before SCM registration |
| `log_file.try_clone()` + `Redirect::stdout` | `.unwrap()` / `.expect` | `.ok().and_then(...)` | Redirect is best-effort; service must run even if log setup partially fails |
| `Redirect::stderr` | `.expect` | `.ok()` | Same — stderr redirect is not essential to service operation |
| `log_path.to_str()` | `.unwrap()` | `to_string_lossy()` | Non-UTF-8 path (rare on Windows) no longer panics |
| `shutdown_tx.send()` (×2) | `.unwrap()` | `let _` | Send fails only if receiver already dropped — that is the normal shutdown case |
| `register(...)` | `.unwrap()` | `.expect("...")` | Registration failure is truly unrecoverable; `expect` at least surfaces the reason in the Windows Application event log |
| `RuntimeBuilder::build()` | `.unwrap()` | `.expect("...")` | Same rationale — unrecoverable, but the message helps diagnosis |

#### `run_service` (returns `windows_service::Result<()>`)

| Site | Old | New | Rationale |
|------|-----|-----|-----------|
| `Nvml::init()` | `.unwrap()` | `match` + log + `return Ok(())` | Let SCM restart policy apply instead of silent panic |
| `Gpu::enumerate()` | `.unwrap()` | `match` + log + `return Ok(())` | Same |
| `start_interval.lock()` | `.unwrap()` | `.unwrap_or_else(\|e\| e.into_inner())` | Recover from poisoned mutex |
| `config.lock()` | `.unwrap()` | `.unwrap_or_else(\|e\| e.into_inner())` | Same |
| `device_by_index(i)` | `.unwrap()` | `match` + log + `continue` | Transient NVML device query failure skips one GPU for this tick |
| `duration_since(UNIX_EPOCH)` | `.unwrap()` | `.unwrap_or_default()` | Clocks going backwards is impossible but not type-enforced |

---

## Relationship to PR #72

PR #72 (`fix(auto-optimizer,srv): GPU selection + path-traversal + srv log/panic/OOB fixes`) already addresses:
- Log directory path (#66) — replaces the `exe.parent().parent()` chain with `%PROGRAMDATA%\nvoc\logs`
- HTTP thread panic hook (#67)
- NVML/NVAPI device-count OOB (#73)

The two `.unwrap()` calls on `env::current_exe()` and `.parent()` (lines 75–79 in the original) are intentionally left here for PR #72 to replace wholesale, avoiding a merge conflict. All other `.unwrap()` calls fixed in this PR are in code sections that PR #72 does not touch.

When both PRs are ready to merge, this one can land first (no overlap in the modified lines) or after #72 with a trivial rebase.

---

## Test plan

- [x] `python3 -m py_compile` — N/A (Rust only)
- [x] `cargo check` — N/A (Windows-only crates; service compiles on Windows CI)
- [ ] Windows: service starts, logs appear in expected path, no startup panic in Event Viewer
- [ ] Windows: stop service via SCM → clean shutdown, no zombie process
- [ ] Windows: simulate `Nvml::init()` failure → service logs error and exits cleanly; SCM marks it stopped
- [ ] Windows: simulate `device_by_index` failure for GPU 1 → only GPU 1 skipped for that tick, service continues
- [ ] Windows: screen size query on GPU with large `dmDriverExtra` → no stack corruption

https://claude.ai/code/session_01PdUm7xHqC2YXCBg91ZDQb6

---
_Generated by [Claude Code](https://claude.ai/code/session_01PdUm7xHqC2YXCBg91ZDQb6)_

### DIFF
--- a/auto-optimizer/src/basic_func.rs
+++ b/auto-optimizer/src/basic_func.rs
@@ -52,11 +52,12 @@ fn get_primary_screen_size_raw() -> (u32, u32) {
     const OFFSET_PELS_WIDTH: usize = 172;
     const OFFSET_PELS_HEIGHT: usize = 176;
     const ENUM_CURRENT_SETTINGS: u32 = 0xFFFF_FFFF;
-    #[repr(C, align(4))]
-    struct AlignedBuf([u8; 256]);
     unsafe {
-        let mut aligned = AlignedBuf([0u8; 256]);
-        let buf = &mut aligned.0;
+        // dmDriverExtra (offset 70) is a u16, so a driver may write up to 65 535 bytes
+        // of private data after the 220-byte DEVMODEW base.  A fixed 256-byte stack buffer
+        // only has 36 bytes of headroom — not guaranteed safe on all driver versions.
+        // Use a heap Vec to cover the full possible range without stack pressure.
+        let mut buf = vec![0u8; DEVMODEW_SIZE + u16::MAX as usize];
         buf[OFFSET_DM_SIZE] = DEVMODEW_SIZE as u8;
         buf[OFFSET_DM_SIZE + 1] = (DEVMODEW_SIZE >> 8) as u8;
         let ret = EnumDisplaySettingsW(std::ptr::null(), ENUM_CURRENT_SETTINGS, buf.as_mut_ptr());
--- a/srv/src/bin/nvoc_service.rs
+++ b/srv/src/bin/nvoc_service.rs
@@ -81,7 +81,9 @@ mod nvoc_service {
             .and_then(|p| p.parent())    // 第二级
             .unwrap_or(&exe_dir)         // 如果失败就用 exe_dir
             .join("logs");                // 加上 logs 目录
-        std::fs::create_dir_all(&log_dir).unwrap();
+        if let Err(e) = std::fs::create_dir_all(&log_dir) {
+            eprintln!("Warning: failed to create log directory {:?}: {}", log_dir, e);
+        }
 
         let log_path = log_dir.join(format!("canbedel-{}-output.log", SERVICE_NAME));
         let log_path_for_log2 = log_path.clone();
@@ -92,13 +94,16 @@ mod nvoc_service {
             .expect("Failed to open log file");
         
         // 重定向 stdout 和 stderr 到同一个文件
-        let _stdout_redirect = Redirect::stdout(log_file.try_clone().unwrap())
-            .expect("Failed to redirect stdout");
-        let _stderr_redirect = Redirect::stderr(log_file)
-            .expect("Failed to redirect stderr");
+        let _stdout_redirect = log_file
+            .try_clone()
+            .ok()
+            .and_then(|f| Redirect::stdout(f).ok());
+        let _stderr_redirect = Redirect::stderr(log_file).ok();
 
         // 2. 初始化 log2（提供轮转和日志级别）
-        let _logger = log2::open(log_path_for_log2.to_str().unwrap())
+        // to_string_lossy handles the (rare) case where the path contains non-UTF-8 bytes.
+        let log_path_str = log_path_for_log2.to_string_lossy();
+        let _logger = log2::open(log_path_str.as_ref())
             .size(100 * 1024 * 1024)  // 100MB
             .rotate(2)                 // 保留1个备份
             // .tee(true)                  // 同时输出到终端
@@ -125,14 +130,16 @@ mod nvoc_service {
 
                 // Handle stop
                 ServiceControl::Stop => {
-                    shutdown_tx.send(()).unwrap();
+                    // Discard send errors: the only failure mode is a dropped receiver,
+                    // which means run_service already exited — that is fine.
+                    let _ = shutdown_tx.send(());
                     ServiceControlHandlerResult::NoError
                 }
 
                 // treat the UserEvent as a stop request
                 ServiceControl::UserEvent(code) => {
                     if code.to_raw() == 130 {
-                        shutdown_tx.send(()).unwrap();
+                        let _ = shutdown_tx.send(());
                     }
                     ServiceControlHandlerResult::NoError
                 }
@@ -143,7 +150,10 @@ mod nvoc_service {
 
         // Register system service event handler.
         // The returned status handle should be used to report service status changes to the system.
-        let status_handle = service_control_handler::register(SERVICE_NAME, event_handler).unwrap();
+        // If SCM registration fails the service cannot report its own state — panic so
+        // the Windows Application event log captures the reason.
+        let status_handle = service_control_handler::register(SERVICE_NAME, event_handler)
+            .expect("Failed to register nvoc_service control handler with SCM");
 
         // Tell the system that service is running
         let _ = status_handle.set_service_status(ServiceStatus {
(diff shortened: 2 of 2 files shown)


Dixiao-L:

PR #98 is the consolidated safety-fixes PR that covers both changes here (DEVMODEW heap buffer → Vec, `nvoc_service.rs` `.unwrap()` removal) along with CSRF guard (#90), gpu-index bound (#91), VFP merge-join (#92), `checked_add` (#93), and voltage range guard (#82). CI passed.

---

Issue #93, opened 2026-05-11
auto-optimizer: unchecked `i32 + i32` on user-CSV `col3_value` in `export_single_point` row-update loop


Dixiao-L:
## Summary

`auto-optimizer/src/oc_profile_function.rs:131-136` updates rows in a user-supplied CSV (`--initcsv` / `--output`) and computes:

```rust
if columns.first().map(String::as_str) == Some(voltage.as_str()) && columns.len() > 3 {
    let col3_value: i32 = columns[3].parse().unwrap_or(0);
    let sum = new_delta + col3_value;            // ← unchecked i32 + i32
    columns[1] = sum.to_string();
    columns[2] = delta;
}
```

`new_delta` is the user's delta input; `col3_value` comes from a CSV field. **Both are `i32`** and the sum has no overflow check:

- In debug builds: panics (`attempt to add with overflow`).
- In release builds (`panic = "abort"` in `Cargo.toml`): silently wraps to `i32::MIN`, writes the wrong value back to the CSV, and the next read interprets that as a real OC delta — i.e. -2.1 GHz when the user intended +2 GHz.

## Reproducer

1. Hand-craft `init.csv` with a row whose 4th column is `2000000000` (close to i32::MAX).
2. `nvoc set vfp --initcsv init.csv --output out.csv --voltage <matching> --delta 200000000`.
3. Inspect `out.csv` column 1 — release: wraps; debug: aborts.

Because the CSV is treated as trusted input but actually comes from `--initcsv` (user-controlled path, even if guarded by #68/PR #72 against `..`), a typo or a stale CSV from a previous session can produce this.

## Distinct from open issues/PRs

- #68 / PR #72 — `--output` / `--initcsv` **path traversal**. That fix rejects `..` paths but does not validate CSV *contents*.
- #74 / PR #72 — `Box::leak` per row. Same function (`export_single_point`), different bug. The Box::leak fix replaces the column type but does **not** add overflow checks on the `+` (the line `columns[1] = sum.to_string()` already exists in PR #72's diff with the same `+`).
- #16 / #82 / PR #87 / PR #84 — `saturating_mul` sweep is only on `oc_get_set_function_nvapi.rs`; `oc_profile_function.rs` was not in scope.

## Suggested fix

```rust
let col3_value: i32 = columns[3].parse().unwrap_or(0);
let sum = new_delta.saturating_add(col3_value);
columns[1] = sum.to_string();
```

[...]

Issue #92, opened 2026-05-11
auto-optimizer: `assert_eq!(i0, i1)` in `collect_vf_points` aborts the process on VFP table/delta index mismatch (release builds)


Dixiao-L:
## Summary

`auto-optimizer/src/oc_profile_function.rs:179` uses `assert_eq!(i0, i1)` to require that NVAPI's VFP-table point index matches the corresponding VFP-delta index. Unlike `debug_assert!`, **`assert!` and `assert_eq!` run in release builds**, and with the workspace's `panic = "abort"` setting (`Cargo.toml:7`) the failure terminates the process with no chance for a Rust panic hook, logger flush, or `Result` upcast.

```rust
fn collect_vf_points(vfp: VfpTable, deltas: VfpDeltas, export_memory: bool) -> Vec<VfPoint> {
    let (points, deltas) = if export_memory { (vfp.memory, deltas.memory) } else { (vfp.graphics, deltas.graphics) };
    points
        .into_iter()
        .zip(deltas)
        .map(|((i0, mut point), (i1, delta))| {
            assert_eq!(i0, i1);              // ← aborts on driver/firmware quirk
            if export_memory {
                point.voltage = Microvolts(point.voltage.0 * 2);
            }
            VfPoint::new(point, delta)
        })
        .collect()
}
```

```toml
# Cargo.toml
[profile.release]
panic = "abort"
```

## Why it matters

This function is called from the V-F-curve export path used by `nvoc set vfp` / `nvoc reset vfp` / autoscan VFP export, all of which now run under the service (with PR #72 merged) as SYSTEM. A driver/firmware update that re-orders VFP indices, a MIG-partitioned datacenter SKU, or a transient NVAPI bug will:

1. Today: abort the process mid-write, with no log line explaining why. Under the service this looks like a silent service exit; under the CLI it looks like a clean exit code != 0 with no `stderr` (because `panic = "abort"` skips Rust's stderr panic message under some PRs of the standard library's signal handling).
2. With graceful handling: return `Error::Custom("VFP index mismatch at point N: table={i0}, delta={i1}")` and let the caller decide whether to fall back to a CSV-only export, retry NVAPI, or warn the user.

Note: this is **separate** from the `* 2` on line 181 (memory voltage doubling), which also lacks `.saturating_mul()` but is bounded by the NVAPI-reported voltage range — a follow-up for the #16/#82 sweep, but lower priority than the unrecoverable abort.

## Distinct from open issues/PRs

[...]

Issue #91, opened 2026-05-11
srv/websrv: `gpu_index` from `?gpu=` query has no upper bound — unbounded `usize` sent into command channel


Dixiao-L:
## Summary

In `srv/src/websrv.rs:160-177` the `/oc_global` handler parses the `gpu` query parameter as `usize` and forwards the raw value into `NVOCServiceCmd { gpu_index, ... }` with no upper bound:

```rust
let gpu_index: usize = match params.get("gpu") {
    None => 0,
    Some(s) => match s.parse::<usize>() {
        Ok(n) => n,
        Err(_) => { /* 400 */ continue; }
    },
};
// ...
cmd_tx.send(NVOCServiceCmd { gpu_index, cmd: "set_oc_global".to_string(), over_freq: freq_val })
```

`?gpu=18446744073709551615` parses cleanly and is enqueued; rejection (or, worse, indexing) happens downstream, far from the input boundary.

## Why it matters

PR #72 adds an `idx >= gpu_dynamic_lock_point.len()` guard in `nvoc_service.rs` for **NVML/NVAPI count drift**, but that guard is in the periodic VFP control loop, not in the command-channel consumer that handles `set_oc_global`. The command-channel consumer historically does `gpus[gpu_index]` or similar — at minimum, a malicious local client (see also the localhost-CSRF issue I'm filing alongside this) can flood the channel with bogus indices, all of which the service has to allocate, queue, log, and then individually fail.

Even with PR #72 merged:
- The HTTP layer is the natural place to reject this: the service knows how many GPUs it enumerated.
- "Reject at the boundary" is a stronger contract than "every downstream consumer must remember to guard".

## Distinct from open issues/PRs

- #73 / PR #72 — NVML/NVAPI count *drift* in the **periodic VFP control loop** (different code path; that fix doesn't extend to the HTTP-driven `set_oc_global` consumer).
- #67 / PR #72 — HTTP panic hook (orthogonal).
- The temperature endpoint range guard (`TEMP_LIMIT_MIN..=TEMP_LIMIT_MAX`) on line 125 and the OC-delta range guard (`OC_DELTA_MIN..=OC_DELTA_MAX`) on line 180 demonstrate the project's own preferred pattern — validate at the HTTP layer. `gpu_index` is the one input that's missed.

## Suggested fix

[...]

Issue #90, opened 2026-05-11
srv/websrv: no HTTP method check on state-mutating endpoints — localhost CSRF from any browser tab


Dixiao-L:
## Summary

`srv/src/websrv.rs` exposes state-mutating endpoints (`/set_temp_limit_soft_vfp`, `/oc_global`) on `127.0.0.1:14514` that match on the URL path only and accept **any** HTTP method, with no `Origin`/`Referer` check, no CSRF token, and no auth. Any local webpage rendered in a browser on the same machine can therefore drive the service — including raising/lowering the OC delta and changing the temp limit — via a single image tag or a no-preflight `fetch()`.

## Evidence

`srv/src/websrv.rs:94-102` matches on `path` only:
```rust
for request in server.incoming_requests() {
    let full_url = request.url().to_owned();
    let (path, query_str) = ...;
    let params = parse_query(query_str);
    match path {
        "/config" => { ... }
        "/set_temp_limit_soft_vfp" => { ... cfg.temp_limit = limit; ... }
        "/oc_global" => { ... cmd_tx.send(NVOCServiceCmd { cmd: "set_oc_global".to_string(), ... }) ... }
        _ => { ... }
    }
}
```

`grep -n "request.method\|Method::" srv/src/websrv.rs` returns nothing — the method is never inspected.

The comment on line 84 ("Bind only to loopback; network-reachable clients cannot reach this endpoint") is correct about remote clients but **does not protect against same-machine browsers** — `127.0.0.1` is reachable from any tab the user has open.

## Attack scenarios

A user with the service installed visits a webpage that contains:

```html
<!-- GET endpoints — no preflight, no CORS check needed by browser -->
<img src="http://127.0.0.1:14514/oc_global?oc=2000000&gpu=0" />
<img src="http://127.0.0.1:14514/set_temp_limit_soft_vfp?limit=120" />
```

Result: the browser fires the GET on render, the service applies a +2 GHz core delta and raises the temp limit to 120 °C. The user is not prompted, no UI is shown, the operation succeeds because the request comes from a local-origin client.

`fetch()` with `mode: "no-cors"` works equivalently for any other method too, since tiny_http processes the request body regardless of method.

## Root cause

State-mutating side effects sitting behind GET, with no method gate and no Origin/Referer validation. This is the textbook localhost-CSRF pattern that affected (e.g.) Transmission's RPC and several mining-pool dashboards.

## Distinct from open issues/PRs

[...]

PR #98, opened 2026-05-11
fix(srv,auto-optimizer): consolidated safety fixes (closes #7, #66, #67, #68, #73, #82, #90, #91, #92, #93)

### LINKED ISSUE #7: srv: replace `.unwrap()` calls with proper error handling in service startup and HTTP loop

Dixiao-L:
## Summary

This PR consolidates the fixes from draft PRs #84, #85, #86, #87, #88, and #94 into one clean commit, resolving the duplicate-PR overlap. Four files, ten issues.

> **Supersedes** (same root causes, better/more complete fix): #84, #85, #86, #87, #88, #94

---

### `srv/src/bin/nvoc_service.rs` — Issues #7, #66, #67, #73

| Site | Before | After | Why |
|------|--------|-------|-----|
| Log dir | `exe.parent().parent().unwrap()` | `%PROGRAMDATA%\nvoc\logs` | SYSTEM-owned path, not writable-install-relative (#66) |
| Log filename | `canbedel-nvoc_service-output.log` | `nvoc_service-output.log` | Remove debug prefix |
| `create_dir_all` | `.unwrap()` | `if let Err(e)` + eprintln | Log dir failure shouldn't abort before SCM registration (#7) |
| `log_file.try_clone()` | `.unwrap()` | `.ok().and_then(...)` | Redirect is best-effort (#7) |
| `log_path.to_str()` | `.unwrap()` | `.to_string_lossy()` | Non-UTF-8 paths won't panic (#7) |
| `shutdown_tx.send()` ×2 | `.unwrap()` | `let _ = ...` | Only fails when receiver already dropped — that is normal shutdown (#7) |
| `register(...)` | `.unwrap()` | `.expect("...")` | Unrecoverable; message surfaces in Windows Application event log (#7) |
| `RuntimeBuilder::build()` | `.unwrap()` | `.expect("...")` | Same (#7) |
| HTTP thread | bare `thread::spawn` | `catch_unwind` + 1 s restart loop | Panic kills the thread silently; service reports Running but control endpoint is gone (#67) |
| `Nvml::init()` | `.unwrap()` | match + log + `return Ok(())` | Let SCM restart policy apply (#7) |
| `Gpu::enumerate()` | `.unwrap()` | match + log + `return Ok(())` | Same (#7) |
| `start_interval` / `config` Mutex | `.unwrap()` | `.unwrap_or_else(\|e\| e.into_inner())` | Recover from poisoned mutex (#7) |
| `device_by_index(i)` | `.unwrap()` | match + log + `continue` | Bad handle for GPU i skips that GPU, others unaffected (#7) |
| `duration_since(UNIX_EPOCH)` | `.unwrap()` | `.unwrap_or_default()` | Clocks-backward is impossible but not type-enforced (#7) |
| VFP control loop | `gpu_dynamic_lock_point[i as usize]` with no bound check | `idx >= gpus.len()` guard + skip | NVML may see more GPUs than NVAPI (MIG, datacenter); OOB panic (#73) |

---

### `auto-optimizer/src/oc_profile_function.rs` — Issues #68, #92, #93

- **`export_single_point` arg lookups**: `.unwrap()` → `ok_or_else(|| Error::Custom(...))?` (#68)
- **`reject_dotdot()`**: new guard rejects any path containing `..` before filesystem access when running as admin (#68)
- **`new_delta + col3_value`**: bare `+` → `.checked_add().ok_or_else(...)` — overflow would silently write a corrupted kHz delta to CSV in release builds with `panic = "abort"` (#93)
- **`assert_eq!(i0, i1)`**: replaced with `filter_map` + `eprintln!` warning — the assert aborts in release builds, killing the process on driver index inconsistency with no log entry (#92)

---

### `srv/src/websrv.rs` — Issues #90, #91

- **`is_mutation_request()`**: new helper requiring POST + `X-Requested-With: XMLHttpRequest`. POST blocks simple-form/GET CSRF; the non-simple header forces a CORS preflight that browsers will refuse cross-origin (#90)
- `/set_temp_limit_soft_vfp` and `/oc_global` both guarded with HTTP 405 when the check fails
- **`GPU_INDEX_MAX = 63`**: any `gpu=` value above this returns HTTP 400 (#91)

---

### `auto-optimizer/src/oc_get_set_function_nvapi.rs` — Issue #82

- `input_voltage * 1000` → `.saturating_mul(1000)`, consistent with every other mV→µV conversion in the file
- Range guard 500 000–2 000 000 µV (0.5–2.0 V) before the NVAPI driver call — rejects e.g. `--voltage 9999` (≈ 10 V) with a clear error message (#82)

---

## Test plan

- [x] `cargo check -p nvoc-auto-optimizer` — zero new errors or warnings
- [x] `cargo check -p nvoc-srv` — zero new errors or warnings
- [ ] Windows: service logs appear in `%PROGRAMDATA%\nvoc\logs\nvoc_service-output.log`
- [ ] Windows: rename `nvml.dll` → service logs "NVML init failed" and SCM shows Stopped, not Running
- [ ] Windows: send SC STOP → clean shutdown, no SCM hang
- [ ] Windows: inject HTTP thread panic → service logs restart, endpoint recovers after 1 s
- [ ] Windows: `nvoc export vfp` with NVML > NVAPI GPU count → warning logged, service continues
- [ ] Windows: craft CSV with `col3 = 2_000_000_000`, `delta = 200_000_000` → error instead of garbage
- [ ] Windows: `nvoc export vfp` with matched indices → same output as before
- [ ] Windows: GET `/oc_global?oc=100000&gpu=0` → HTTP 405
- [ ] Windows: POST `/oc_global` without `X-Requested-With` → HTTP 405
- [ ] Windows: POST `/oc_global` with `X-Requested-With: XMLHttpRequest` → HTTP 200
- [ ] Windows: POST `/oc_global?oc=100000&gpu=99` → HTTP 400
- [ ] Windows: `nvoc set nvapi --voltage 9999` → rejected with range message
- [ ] Windows: `nvoc set nvapi --voltage 1000` → accepted (1 000 000 µV in range)

https://claude.ai/code/session_01BgpAaj8QMKA6shvgjKJMmk

---
_Generated by [Claude Code](https://claude.ai/code/session_01BgpAaj8QMKA6shvgjKJMmk)_


Dixiao-L:
CI triage (2026-05-12):

- `srv (windows)` ✅ — 3 min 
- `auto-optimizer (windows)` ✅ — 2.5 min
- All other jobs correctly skipped (no Python/stressor files changed)

Code review notes:
- **Merge-join** in `collect_vf_points` is correct and directly resolves the review-thread concern raised on PR #94 (the `zip()` silent truncation after the first index mismatch).
- **CSRF guard** (`is_mutation_request`) is correct: POST + non-simple `X-Requested-With` forces a CORS preflight for all cross-origin browser tabs.
- **Voltage range guard** and **`saturating_mul`** are both correct and consistent with the rest of the file.
- **OOB guard** for `gpu_dynamic_lock_point` correctly uses `gpus.len()` as the bound.
- **Log dir** moved to `%PROGRAMDATA%\nvoc\logs` — right call for a SYSTEM-owned service.
- All error paths look correct. No remaining panic-prone `.unwrap()` calls in hot paths.

One heads-up: the base is `9db72d2` (before PR #96 landed on main). There are no conflicts with the Rust files, but a "Update branch" click will be needed once the `gpu-ci.yml` revert (PR #102 / PR #105) lands. The Rust changes here are entirely independent of the CI yaml.

---

physicsdolphin (dismissed):

physicsdolphin (dismissed):

PR #99, opened 2026-05-11
style(fmt): force formatting with rustfmt.toml

adamanteye:


physicsdolphin:
postpone this until agentic-overhaul is all combined

PR #102, opened 2026-05-12
Revert "Update gpu-ci.yml"

physicsdolphin:
Reverts Skyworks-Neo/nvoc#96

runner.* context：

只能在 step 里用
不能在 job-level env: 里用

还是写死E盘得了，Windows目前就一台CI runner，固定即可，先跑起来再说

### DIFF
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -68,12 +68,10 @@ jobs:
             stressor_cuda:
               - 'cli-stressor-cuda/**'
               - 'pyproject.toml'
-              - 'uv.lock'
               - '.github/workflows/gpu-ci.yml'
             stressor_opencl:
               - 'cli-stressor-opencl/**'
               - 'pyproject.toml'
-              - 'uv.lock'
               - '.github/workflows/gpu-ci.yml'
 
   auto-optimizer-gpu:
@@ -105,9 +103,6 @@ jobs:
     if: needs.changes.outputs.stressor_cuda == 'true'
     runs-on: [self-hosted, windows, gpu-nvidia, cuda]
     timeout-minutes: 30
-    env:
-      UV_CACHE_DIR: ${{ runner.temp }}\uv-cache
-      UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}\uv-envs\cuda-stressor-${{ github.run_id }}
     defaults:
       run:
         working-directory: cli-stressor-cuda
@@ -116,7 +111,7 @@ jobs:
         with:
           ref: ${{ inputs.ref || github.ref }}
       - uses: astral-sh/setup-uv@v4
-      - run: uv sync --frozen
+      - run: uv sync
       - name: short stress run
         run: uv run python test.py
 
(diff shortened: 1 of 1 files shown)


physicsdolphin:
思考，我应该revert合并恢复分支，而不是新开一个分支反向revert

Dixiao-L (approved):
Approved — the root cause is correct: `runner.*` context expressions are evaluated lazily and are only valid inside a `step`, not in a `job`-level `env:` block. GitHub Actions simply produces an empty string (or a literal) for them there, which caused the job to hang waiting on an invalid path.

The revert unblocks CI. Two things were dropped along the way that are worth restoring in a follow-up:

**1. `uv.lock` removed from path filters**
The lockfile lives at the workspace root and pins every stressor dependency. If someone merges a `uv.lock` update (e.g. a security bump) without touching `cli-stressor-cuda/**` or `cli-stressor-opencl/**`, GPU CI will not run and the update goes untested on the GPU runner. Suggest adding it back:
```yaml
stressor_cuda:
  - 'cli-stressor-cuda/**'
  - 'pyproject.toml'
  - 'uv.lock'          # <-- restore
  - '.github/workflows/gpu-ci.yml'
```

**2. `uv sync --frozen` → `uv sync`**
`--frozen` refuses to update the lockfile and fails fast if the environment diverges from `uv.lock`. Without it, a stale or incompatible lock can silently get patched on the runner, making CI results non-reproducible across runs.

**Proper fix for the original `runner.temp` issue** (avoids needing the E-drive hardcode): move the `UV_CACHE_DIR` / `UV_PROJECT_ENVIRONMENT` env vars from job level to the step level, where `runner.*` context is valid:
```yaml
      - run: uv sync --frozen
        env:
          UV_CACHE_DIR: ${{ runner.temp }}\uv-cache
          UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}\uv-envs\cuda-stressor-${{ github.run_id }}
```
This keeps per-run venv isolation, preserves `--frozen`, and requires no hardcoded paths. Happy to send a follow-up PR with this change once this revert is merged.

---

Dixiao-L (commented):
## Review: Revert of PR #96

The root cause is correct: GitHub Actions does **not** interpolate the `runner` context at the `jobs.<job>.env` level — only inside `steps`. The env block was silently treated as literal strings, which caused `uv` to try to create directories named `${{ runner.temp }}\uv-cache` literally on disk, hanging or failing on the Windows self-hosted runner.

Removing those env vars and using uv's default cache location is the cleanest fix.

---

### One observation: `uv.lock` removed from change filters

PR #96 added `uv.lock` to the stressor filter lists so that a dependency-version bump (lockfile update) would automatically trigger the GPU stressor tests. This revert removes those entries:

```yaml
stressor_cuda:
  - 'cli-stressor-cuda/**'
  - 'pyproject.toml'
- - 'uv.lock'         # ← removed
  - '.github/workflows/gpu-ci.yml'
```

This is a minor regression: a `uv.lock`-only change (e.g. a transitive dependency update) will no longer trigger GPU CI automatically. Given that GPU CI is already gated behind the `ci:gpu` label and manual dispatch, this is acceptable, but worth noting. You can restore just the `uv.lock` filter line independently when the CI is stable, without re-introducing the broken env vars.

---

**Fix is correct** — approving the revert as-is. The `--frozen` removal (`uv sync` instead of `uv sync --frozen`) is also appropriate for a single dedicated runner where lock-sync failures are more harmful than flexibility.

---

PR #109, opened 2026-05-13
fix(clippy): remove useless format! wrappers around static strings

Dixiao-L:
## Summary

- Replace two `format!("NVML backend unavailable")` calls (no format args) in `auto-optimizer/src/main.rs` with plain string literals — fixes `clippy::useless_format` at lines 164 and 172.

## Root cause

`format!` with a static string and no interpolation args is a no-op allocation; clippy flags it as `useless_format`. Since the error type is `Box<dyn std::error::Error>`, `.into()` works directly on `&str` via `impl From<&str> for Box<dyn Error>`.

## Fix

```rust
// before
return Err(format!("NVML backend unavailable").into());

// after
return Err("NVML backend unavailable".into());
```

## Test plan

- [ ] `cargo clippy -p nvoc-auto-optimizer` — no `useless_format` warnings on these lines
- [ ] `cargo check -p nvoc-auto-optimizer` — clean

https://claude.ai/code/session_015c8BLtJS3Cb1GDvYzWBRBa

PR #110, opened 2026-05-14
fix(auto-optimizer): saturating_add for i32 delta sum in export_single_point

### LINKED ISSUE #93: auto-optimizer: unchecked `i32 + i32` on user-CSV `col3_value` in `export_single_point` row-update loop

Dixiao-L:
## Summary

- Replace `y_value + col3_value` with `y_value.saturating_add(col3_value)` in `auto-optimizer/src/oc_profile_function.rs:152` — closes #93.

## Root cause

`export_single_point` updates rows in a user-supplied CSV. It sums the new OC delta (`y_value`, parsed from the just-written `parts[2]`) with a pre-existing column value (`col3_value`, from `parts[3]`). Both are `i32`. With `panic = "abort"` in release builds, an unchecked `i32 + i32` overflow wraps to `i32::MIN` and writes the corrupted value back to the CSV — the next read sees it as a ~−2.1 GHz delta when the user intended +200 MHz.

## Fix

```rust
// before
parts[1] = (y_value + col3_value).to_string();

// after
parts[1] = y_value.saturating_add(col3_value).to_string();
```

`saturating_add` clamps to `i32::MAX`/`i32::MIN` instead of wrapping, matching the project's established style from the `saturating_mul` sweep in PR #57 / PR #87 (`oc_get_set_function_nvapi.rs`).

## Test plan

- [ ] `cargo check -p nvoc-auto-optimizer` — clean
- [ ] Normal small deltas (e.g. `--delta 100000`) — output unchanged
- [ ] CSV with `col3_value = 2_000_000_000` + `--delta 200_000_000`: before = wraps to negative; after = clamps to `i32::MAX`

https://claude.ai/code/session_015c8BLtJS3Cb1GDvYzWBRBa

### DIFF
--- a/auto-optimizer/src/oc_profile_function.rs
+++ b/auto-optimizer/src/oc_profile_function.rs
@@ -149,7 +149,7 @@ pub fn export_single_point(point: VfPoint, matches: &clap::ArgMatches) -> Result
             parts[2] = delta_str.clone();
             let y_value: i32 = parts[2].parse().unwrap_or(0);
             let col3_value: i32 = parts[3].parse().unwrap_or(0);
-            parts[1] = (y_value + col3_value).to_string();
+            parts[1] = y_value.saturating_add(col3_value).to_string();
         }
         record_lines.push(parts.join(","));
     }


Dixiao-L:
Superseded by #111, which consolidates this fix alongside the `useless_format` clippy fix from #109 into a single auto-optimizer PR. Closing this one to keep the PR list clean.

---

PR #111, opened 2026-05-14
fix(auto-optimizer): saturating_add overflow + remove useless format! wrappers

Dixiao-L:
## Summary

Two related `auto-optimizer` fixes consolidated from PRs #109 and #110:

- **`oc_profile_function.rs:152`** — replace `y_value + col3_value` with `y_value.saturating_add(col3_value)` to prevent i32 overflow corrupting the CSV (closes #93)
- **`main.rs:164,172`** — replace `format!("NVML backend unavailable")` (no args) with plain `"NVML backend unavailable"` string literal — eliminates `clippy::useless_format` (closes #109)

## Root causes

**Overflow (#93):** `export_single_point` sums new OC delta (`y_value`) and pre-existing column value (`col3_value`). Both are `i32`. With `panic = "abort"` in release builds, unchecked `i32 + i32` overflow wraps to `i32::MIN`, writing a ~−2.1 GHz delta back to the CSV on the next read. `saturating_add` clamps to `i32::MAX`/`i32::MIN` instead, matching the `saturating_mul` pattern from PR #57/#87.

**Clippy lint:** `format!` with a static string and no interpolation args allocates needlessly. `&str` already implements `Into<Box<dyn Error>>` so `.into()` works directly.

## Test plan

- [x] `cargo check -p nvoc-auto-optimizer` — clean
- [x] `cargo clippy -p nvoc-auto-optimizer` — no `useless_format` warnings
- [ ] Normal small deltas (e.g. `--delta 100000`) — output unchanged
- [ ] CSV with `col3_value = 2_000_000_000` + `--delta 200_000_000`: before wraps negative; after clamps to `i32::MAX`

Supersedes #109 and #110.

https://claude.ai/code/session_01K6F2hWLxxwtvae9qv9vPBK

---
_Generated by [Claude Code](https://claude.ai/code/session_01K6F2hWLxxwtvae9qv9vPBK)_

PR #113, opened 2026-05-14
fix(stressors): break precision loop on GPU fault to prevent Linux hang

### LINKED ISSUE #35: cli-stressor-cuda: `sys.exit(1)` from inside inner loop kills multi-precision sweep before summary

Dixiao-L:
## Summary

- Both `cli-stressor-cuda` and `cli-stressor-opencl` now break out of the per-precision loop immediately after a real (non-SKIP) failure — closes #35.

## Root cause

The inner stress loop was fixed in PR #59 to use `break` instead of `sys.exit(1)`, which preserved multi-precision summaries. However the outer `for spec in precisions` loop in `main()` had no corresponding guard — after one precision failed, `main()` silently continued dispatching GEMM kernels to the next precision.

On Linux, a GPU in an OC-fault state does not self-recover (no TDR equivalent). Firing another kernel into a fallen-off-bus device risks a permanently stuck process that blocks the autoscan recovery script and may require a reboot. @physicsdolphin flagged this in issue #35.

## Fix

```python
# after printing the per-precision result:
if res.first_error and res.supported:   # CUDA
    break

if result_status(res) == "FAIL":        # OpenCL (uses result_status helper)
    break
```

`print_summary()` is still called with all results collected so far, so partial summaries are preserved. The exit code is still non-zero on failure. SKIP results (unsupported precision) do not trigger the break.

> **Note:** the first commit on this branch (`6162e9d`, `saturating_add`) is also being reviewed in PR #111. That commit's diff will disappear from this PR once #111 merges to `main`.

## Test plan

- [ ] CUDA: run with `--precisions fp16,fp32` against a config where FP16 fails; verify FP32 test is **not** started and partial summary is printed
- [ ] OpenCL: same test with two precision modes
- [ ] Normal run (no failures): all precisions complete as before
- [ ] SKIP precision (unsupported): does not trigger early exit; remaining precisions run

https://claude.ai/code/session_01FRS8ASEedS6moqcCWN1Mmj

---
_Generated by [Claude Code](https://claude.ai/code/session_01FRS8ASEedS6moqcCWN1Mmj)_

### DIFF
--- a/auto-optimizer/src/oc_profile_function.rs
+++ b/auto-optimizer/src/oc_profile_function.rs
@@ -149,7 +149,7 @@ pub fn export_single_point(point: VfPoint, matches: &clap::ArgMatches) -> Result
             parts[2] = delta_str.clone();
             let y_value: i32 = parts[2].parse().unwrap_or(0);
             let col3_value: i32 = parts[3].parse().unwrap_or(0);
-            parts[1] = (y_value + col3_value).to_string();
+            parts[1] = y_value.saturating_add(col3_value).to_string();
         }
         record_lines.push(parts.join(","));
     }
--- a/cli-stressor-cuda/test.py
+++ b/cli-stressor-cuda/test.py
@@ -554,6 +554,10 @@ def main():
         else:
             print(f"  结果: completed without detected error")
         print(f"  累计: {res.iterations} 次 matmul, {res.tflops:.2f} TFLOPS")
+        # On Linux a faulted GPU won't recover without a reboot; stop further
+        # precision tests immediately rather than issuing more GEMM kernels.
+        if res.first_error and res.supported:
+            break
 
     print_summary(device_name, total_memory_gb, results)
 
--- a/cli-stressor-opencl/test.py
+++ b/cli-stressor-opencl/test.py
@@ -979,6 +979,10 @@ def main():
         else:
             print("  结果: completed without detected error")
         print(f"  累计: {res.iterations} 次 matmul, {res.tflops:.2f} TFLOPS")
+        # On Linux a faulted GPU won't recover without a reboot; stop further
+        # precision tests immediately rather than issuing more kernels.
+        if result_status(res) == "FAIL":
+            break
 
     print_summary(runtime, results)
     return 1 if any(result_status(result) == "FAIL" for result in results) else 0


Dixiao-L:
Closing in favour of PR #115, which consolidates the stressor fail-fast fix (this PR) with the `saturating_add` overflow fix and `format!` cleanup into one unit. The reviewers already set here (`@physicsdolphin`, `@Master-Hash`) have been requested on #115 as well.

---

Dixiao-L:
This PR is superseded by #115, which consolidates the stressor fail-fast fix (#35) together with the `saturating_add` overflow fix (#93) and the `format!` lint cleanup (#109) into a single PR. Closing in favour of #115.

---

PR #121, opened 2026-05-15
fix(clippy): resolve all warnings and enable -D warnings in CI

Dixiao-L:
## Summary

Addresses the warning backlog raised in PR #119 comment — **enables `-D warnings`** in the `enforce-fmt-lint` CI job.

Builds on the CI enforcement workflow from PR #119 and the `useless_format` fix from PR #118.

### What was wrong (root causes)

| Warning | Root cause | Fix |
|---------|-----------|-----|
| `dead_code` in `srv/src/websrv.rs` (11 items) | All items are only referenced inside `#[cfg(windows)] mod nvoc_service`, so Clippy on Linux sees them as unused | `#[cfg_attr(not(windows), allow(dead_code))]` at top of `websrv.rs` |
| `dead_code` — `GpuSelector::all` / `from_specs` | Intentional future public API with no current caller | `#[allow(dead_code)]` on each function |
| `type_complexity` (4 functions) | Complex tuple return types across nvapi/nvml/profile layers — need `type` alias refactor (design decision) | `#[allow(clippy::type_complexity)]` at each function |
| `too_many_arguments` (4 functions in `oc_scanner.rs`) | Domain-driven argument count in private phase-runner helpers — API refactor out of scope | `#[allow(clippy::too_many_arguments)]` at each function |
| 26 auto-optimizer lints (`collapsible_if`, `match_single_binding`, `needless_late_init`, etc.) | Mechanical style issues | Cherry-picked fixes from PR #118 |
| `useless_format` (2×) | `format!("static string")` with no args | Already fixed in the cherry-picked commit |

### Supersedes

- Draft PR #116 — same auto-optimizer warning sweep (overlapping changes)
- Draft PR #118 — same auto-optimizer warning sweep (cherry-picked here)

Both can be closed once this lands.

## Test plan

- [x] `cargo clippy --all-targets --all-features -- -D warnings` — **zero warnings/errors**
- [x] `cargo fmt --all -- --check` — clean
- [ ] CI enforce-fmt-lint workflow — should pass on all four jobs including Clippy with `-D warnings`
- [ ] CI main build (srv windows, auto-optimizer windows) — no logic changes, should pass

https://claude.ai/code/session_01D5Xpc8UxLgfufmwFVzw9EY

---
_Generated by [Claude Code](https://claude.ai/code/session_01D5Xpc8UxLgfufmwFVzw9EY)_

PR #120, opened 2026-05-15
fix(srv,auto-optimizer,ci): eliminate 15 clippy warnings + clean ruff commands

Dixiao-L:
## Summary

Three targeted fixes that eliminate 15 clippy warnings (50 → 35) and clean up the new `enforce-fmt-lint.yml` workflow.

### `srv`: `#[cfg(windows)]` on `mod websrv` (−13 dead-code warnings)

`websrv.rs` was always compiled via `#[path = "../websrv.rs"] mod websrv;` outside any cfg-gate, but every caller lives inside `#[cfg(windows)]`. On Linux CI this produced 13 spurious `dead_code` warnings:
- `NVOCServiceConfig`, `NVOCServiceCmd` structs
- `start_http_server`, `respond`, `percent_decode`, `parse_query`, `json_content_type`, `is_mutation_request` functions
- `TEMP_LIMIT_MIN/MAX`, `OC_DELTA_MIN/MAX`, `GPU_INDEX_MAX` constants

Fix: add `#[cfg(windows)]` to the `mod websrv;` declaration.

### `auto-optimizer/src/main.rs`: `format!` → `.to_string()` (−2 `useless_format` warnings)

Lines 164 and 172 had `format!("NVML backend unavailable")` which allocates needlessly — a static string literal with `.to_string()` (or the existing `.into()`) is equivalent and avoids the lint.

### `enforce-fmt-lint.yml`: remove duplicate `.` positional arg

Both ruff commands had the current directory passed twice:
```
ruff format . --check --output-format=github .  # duplicate '.'
ruff check . --output-format=github .            # duplicate '.'
```
Removed the redundant first `.`.

## Remaining warnings (35, all auto-optimizer)

| Lint | Count | Status |
|------|-------|--------|
| `collapsible_if` | 13 | addressed in draft PRs #116/#118 |
| `type_complexity` | 4 | deferred — need type-alias design |
| `too_many_arguments` | 4 | deferred — need parameter-struct refactor |
| `needless_late_init` | 4 | addressed in draft PRs #116/#118 |
| `match_single_binding` | 2 | addressed in draft PRs #116/#118 |
| `field_reassign_with_default` | 2 | addressed in draft PRs #116/#118 |
| others (match, borrow, loop) | 4 | addressed in draft PRs #116/#118 |
| `dead_code` (intentional API) | 1 | `GpuSelector::all/from_specs` — intentional |
| `enum_variant_names` | 1 | addressed in draft PRs #116/#118 |

## Test plan

[...]

PR #122, opened 2026-05-15
ci: fix typo `.`

Master-Hash:


Dixiao-L (approved):

Dixiao-L (approved):
Fix is correct and clean. Both ruff commands had `.` passed twice as the target-directory positional argument — once before `--check`/`--output-format` and once at the end. Removing the trailing `.` resolves the duplication without changing the scan scope.

Note: draft PRs #120 and #121 also contain a fix for this same duplicate-`.` issue (using a slightly different form — they move the `.` to the end instead of keeping it in front). Once this PR merges those two drafts will need a rebase/cherry-pick to drop their overlapping hunk; the rest of their changes are independent and still valuable.

---

Dixiao-L:
Fix is correct and minimal. Removing the duplicate `.` positional argument from both ruff commands is the right call — the trailing `.` was redundant (current directory was already specified before the flags) and ruff simply treated it as a second input path, which is harmless but confusing.

CI status:
- ruff format ✅
- ruff check ✅
- Rustfmt ✅
- Clippy (in-progress — only touches YAML so will pass)

**Heads-up on overlap:** PR #120 has a conflicting fix for the same two lines (it moves `.` to the trailing position instead of removing it). Once this merges, #120 will need to rebase on updated main and drop or adapt its CI section. The other changes in #120 (auto-optimizer `format!` fix, `#[cfg(windows)]` on `mod websrv`) are independent and still valid.

LGTM — approving.

---

PR #203, opened 2026-06-06
core, cli-common: Strip nvoc- prefix

84634E1A607A:


Master-Hash (approved):
I dislike it, but anyway, I recognize it.

PR #267, opened 2026-08-19
Umbrella: split thermal, power, and sensor reversing

G4614:
This draft PR is the umbrella and historical integration branch for the thermal, power, sensor, and adjacent reverse-engineering work. It is intentionally **not** intended to merge as one unit.

## Landed in `main`

- [x] #271 — NVML acoustic target temperature
- [x] #272 — NVAPI initialization on older drivers
- [x] #273 — full CLI help on zero-argument invocation
- [x] #274 — `get-status` units and sensor display
- [x] #275 — RTSS thermal channel layout
- [x] #282 — expanded TUI live GPU metrics
- [x] #283 — core, hotspot, and memory temperatures in the TUI
- [x] #285 — legacy NVAPI three-sensor thermal query
- [x] #289 — decoded performance-limit reasons across CLI, Python, and TUI
- [x] #290 — effective GPU clock monitoring
- [x] #291 — read-only PCIe generation, bandwidth, and replay telemetry
- [x] #292 — live NVML board-power draw in CLI and Python
- [x] #293 — cap GUI BLAS thread pools during startup
- [x] #294 — release the Python GIL during read-only `pynvoc` queries
- [x] #295 — cap TUI BLAS thread pools during startup
- [x] #296 — serialize TUI read-only query workers
- [x] #297 — keep TUI V/F refresh data in memory
- [x] #298 — GUI recovery from GPU hot-add and removal
- [x] #299 — TUI recovery from GPU hot-add and removal
- [x] #300 — pause GUI dashboard polling while the window is in the tray
- [x] #301 — deduplicate GUI global scrollwheel handlers
- [x] #302 — portable GUI paths and native fixed fonts
- [x] #303 — remove the production GUI memory-debug sampler
- [x] #304 — cancel GUI CLI processes off the UI thread
- [x] #305 — debounce and serialize GUI config writes
- [x] #306 — batch GUI console output updates
- [x] #307 — pause dashboard polling while another tab is visible
- [x] #308 — skip unchanged GUI slider redraws
- [x] #309 — coalesce post-native-action refresh chains
- [x] #310 — avoid callback reentry during initial GUI layout
- [x] #311 — skip unchanged VF curve redraws
- [x] #312 — organize the reverse-engineering documentation archive

## Active focused child PRs

None. All selected focused child PRs have landed in `main`.

Each selected NVOC PR targeted `main` directly and was reviewed and tested as an independent unit. This avoided another intermediate-branch merge collapsing multiple review units.

[...]

PR #271, opened 2026-08-25
feat(cli): add NVML acoustic target temperature

84634E1A607A:
Extracted from #267 as a single-function PR.

Adds `set-acoustic-temp-c` for NVML `ACOUSTIC_CURR`, including the core operation and CLI wiring. This remains NVML-only and does not alter the existing thermal-limit command.

The branch is rebased onto current `main` and remains draft because the feature is a GPU-mutating target-temperature setter; no write test was needed or executed during this split.

Validation on the rebased head:
- `cargo fmt --all -- --check`
- `cargo build --workspace --exclude cli-stressor-cuda-rs`
- `cargo test --package nvoc-core --all-targets`

No GPU writes were executed.


physicsdolphin:
seems not working on real driver on both windows/linux

PR #279, opened 2026-08-25
feat(monitoring): add read-only PCIe telemetry

84634E1A607A:
Extracted from #267 as a focused read-only PCIe telemetry PR.

Depends on #278 (`split/perf-limit-reasons`) so the dashboard formatter already has the expanded metrics and decoded performance-limit row.

Adds one best-effort NVML telemetry query for current/max PCIe generation, TX/RX throughput, and the cumulative replay counter. CLI and Python status omit unsupported fields; the TUI renders available values on the existing PCIe row and only warns when the replay counter is nonzero.

This intentionally excludes NVAPI clock-domain changes, effective clocks, private PCIe counters, and all GPU writes.

Validation:
- `cargo fmt --all -- --check`
- `ruff format . --preview --check` and `ruff check .`
- `cargo test --package nvoc-core --all-targets`
- `cargo test --package nvoc-cli --all-targets` (35 passed)
- `cargo check --package pynvoc`
- `cargo clippy --workspace --exclude cli-stressor-cuda-rs --all-targets -- -D warnings`
- `cd tui && uv run --python 3.13 pytest` (52 passed)
- elevated read-only `nvml_pcie_telemetry_ok`: passed on the local GPU

No GPU writes were executed.


physicsdolphin:

PR #285, opened 2026-08-25
feat(cli): add legacy thermal settings query

84634E1A607A:
Replacement for #280, which was merged into its intermediate base branch rather than into `main`.\n\nExtracted from #267 as a focused read-only legacy thermal query.\n\nDepends on #275 (`split/thermal-channels`). The RTSS channel view in #275 remains the live-status source; this command exposes the separate standard NVAPI three-sensor view because it also carries each sensor physical minimum and maximum range.\n\nAdds `get-thermal-settings` for GPU, memory, and board readings, with structured core output and compact CLI rendering. It adds no target-temperature policy, simulation control, or setter.\n\nValidation after rebuilding on current `main`:\n- `cargo fmt --all -- --check`\n- `cargo build --workspace --exclude cli-stressor-cuda-rs`\n- `cargo test --package nvoc-core --all-targets`\n- the hardware-gated read-only range test passed before the rebuild and passed on the rebuilt head\n\nNo GPU writes were executed.


physicsdolphin:

physicsdolphin (approved):

PR #286, opened 2026-08-25
feat(monitoring): add read-only PCIe telemetry

84634E1A607A:
Replacement for #279, which was merged into its intermediate base branch rather than into `main`.\n\nExtracted from #267 as a focused read-only PCIe telemetry PR.\n\nDepends on #284 (`split/perf-limit-reasons-v2`).\n\nAdds one best-effort NVML telemetry query for current/max PCIe generation, TX/RX throughput, and the cumulative replay counter. CLI and Python status omit unsupported fields; the TUI renders available values on the existing PCIe row and warns only when the replay counter is nonzero. Effective clocks, private PCIe counters, and all GPU writes remain excluded.\n\nValidation after rebuilding on current `main`:\n- `cargo fmt --all -- --check`\n- `cargo build --workspace --exclude cli-stressor-cuda-rs`\n- `cargo test --package nvoc-core --all-targets`\n- covered by the stack-tip TUI suite (`54 passed`)\n- the hardware-gated read-only PCIe test passed before the rebuild and passed on the rebuilt head\n\nNo GPU writes were executed.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -1444,7 +1444,30 @@ fn execute_target(
             let info = run(target, QueryGpuInfo)?.output;
             Ok(Value::String(info.uuid.unwrap_or_default()))
         }
-        Command::GetStatus => Ok(serde_json::to_value(run(target, QueryGpuStatus)?.output)?),
+        Command::GetStatus => {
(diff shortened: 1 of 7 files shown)


physicsdolphin:

